PR #21, opened 2024-08-28
Ensure no temporary directories are left behind

### LINKED ISSUE #5: grabit v0.1.2 download leaves a temporary directory
## Description

After running `grabit download`, a temporary directory is left over.

## Expected Behavior

Upon a successful `grabit download`, I expect there to be no temporary artifacts left in the filesystem.

## Actual Behavior

A `.blahblahblah` directory is left over in the directory where `grabit` downloads.

## Affected Version

0.1.2

## Steps to Reproduce

```
[root@997607b55ffe testing]# rm -rf foo
[root@997607b55ffe testing]# mkdir foo
[root@997607b55ffe testing]# grabit download --dir foo
[root@997607b55ffe testing]# ls -la foo
total 1208
drwx------ 3 root root   4096 Jul 25 19:11 .
drwx------ 3 root root   4096 Jul 25 19:11 ..
drwx------ 2 root root   4096 Jul 25 19:11 .9Oo3HMo9Iyuqxrw9myzYBObnF
-rw------- 1 root root 168532 Jul 25 19:11 logrotate-3.21.0.tar.xz
[root@997607b55ffe testing]# 
```

The `.9Oo3HMo9Iyuqxrw9myzYBObnF` must have been created by `grabit`.

## Checklist

- [x] I have read the [contributing guidelines](/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

rabadin:
The core problem is that base64 encoded strings can have slashes in them... so this would create intermediate directories that would be left behind. Using a different encoding for the temporary files solves this problem.

Fixes #5

### DIFF
--- a/internal/resource.go
+++ b/internal/resource.go
@@ -6,7 +6,7 @@ package internal
 import (
 	"context"
 	"crypto/sha256"
-	b64 "encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"net/url"
 	"os"
@@ -68,7 +68,7 @@ func GetUrlToDir(u string, targetDir string, ctx context.Context) (string, error
 	// create temporary name in the target directory.
 	h := sha256.New()
 	h.Write([]byte(u))
-	fileName := filepath.Join(targetDir, fmt.Sprintf(".%s", b64.StdEncoding.EncodeToString(h.Sum(nil))))
+	fileName := filepath.Join(targetDir, fmt.Sprintf(".%s", hex.EncodeToString(h.Sum(nil))))
 	return getUrl(u, fileName, ctx)
 }
 


internal/resource.go
```diff
@@ -92,6 +92,10 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 		// Download file in the target directory so that the call to
 		// os.Rename is atomic.
 		lpath, err := GetUrlToDir(u, dir, ctx)
+		defer func(path string) {
+			pathDir := filepath.Dir(path)
+			os.Remove(pathDir)
```
jsquyres, line 97: Is there ever a case will this need to be `os.RemoveAll()`?

rabadin: That's a good idea. In case of error there could be leftovers in there. Changed.

jsquyres (approved):
❤️

PR #22, opened 2024-09-06
Add more unit-tests

rabadin:
This branch's aim is to improve the unit tests to get to > 80% test coverage.

The only non-test / mechanical change is in the download logic: this was changed to print all the errors encountered when failing to download multiple assets as opposed to just the first one. This is in `internal/lock.go`.


cmd/dowload_test.go
```diff
@@ -0,0 +1,119 @@
...
+	[[Resource]]
+	Urls = ['http://localhost:%d/test.html']
+	Integrity = 'sha256-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE='
+	Tags = ['tag']
+
+	[[Resource]]
+	Urls = ['http://localhost:%d/test2.html']
+	Integrity = 'sha256-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE='
+	Tags = ['tag1', 'tag2']
+`, port, port))
```
jsquyres, line 87: Is there an easy way to share the code between here and `TestRunDownloadWithTags()`?  It seems largely the same.

rabadin: Good point: I refactored the tests here to add a few utilities that remove most of the duplication (`test.TestHttpHandler` and `test.AssertFileContains`).
For the test data: I'd rather keep the slight duplication to make sure that each test is completely self-contained: each test can be read and understood (preparation, execution, test) without having to look at utility methods to understand what's happening.

Makefile
```diff
@@ -33,5 +33,9 @@ test: dep
 	@go test -race -v -coverpkg=./... -coverprofile=coverage.out ./...
 	@go tool cover -func coverage.out | tail -n 1
 
+test-gst: dep
+	@gotestsum --format testdox -- -race -coverpkg=./... -coverprofile=coverage.out ./...
```
jsquyres, line 37: Dumb question: where does one get `gotestsum`?

cmd/dowload_test.go
```diff
@@ -2,44 +2,37 @@ package cmd
...
+	port := test.TestHttpHandler(content, t)
 	testfilepath := test.TmpFile(t, fmt.Sprintf(`
 	[[Resource]]
 	Urls = ['http://localhost:%d/test.html']
 	Integrity = 'sha256-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE='
-`, port))
+
+	[[Resource]]
+	Urls = ['http://localhost:%d/test3.html']
+	Integrity = 'sha256-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE='
```
jsquyres, line 21: A very minor point: Is there a way to not have to hard-code this sha256 sum?  E.g., could you compute it dynamically from the `content`?  This is obviously not a *problem* -- just might make it a little easier to have arbitrary content in the tests (vs. the human having to generate the content and matching sha256 sum).

jsquyres (approved):

jsquyres (approved):
❤️

PR #24, opened 2024-09-11
Progress bar (status line)

MDylanWilliams:
The download command has a new flag: --status (-s for shorthand)

Upon executing the download command with this new flag, a line will appear that displays updating metrics. 
Examples: 
      -[___________]          1/1000 Resources          6 B / 6.0 kB          0s elapsed
   ✔[██████]          1/1 Resources          6 B / 6 B          0s elapsed

- The user is shown how many resources have completed downloading (numerically with "#/# Resources" and visually with the bar), how many bytes have downloaded and the total size of all resources, and how much time has elapsed.
- The byte counter, resource counter, and bar update when a resource finishes downloading.
- The ASCII spinner and seconds counter continuously update.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -15,6 +15,8 @@ func addDownload(cmd *cobra.Command) {
 		Args:  cobra.NoArgs,
 		RunE:  runFetch,
 	}
+	downloadCmd.Flags().BoolP("status", "s", false, "Continuously display bytes/resources downloaded, time elapsed, and progress bar")
+	downloadCmd.Flags().Lookup("status").NoOptDefVal = "true"
 	downloadCmd.Flags().String("dir", ".", "Target directory where to store the files")
 	downloadCmd.Flags().StringArray("tag", []string{}, "Only download the resources with the given tag")
 	downloadCmd.Flags().StringArray("notag", []string{}, "Only download the resources without the given tag")
@@ -47,7 +49,11 @@ func runFetch(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	err = lock.Download(dir, tags, notags, perm)
+	status, err := cmd.Flags().GetBool("status")
+	if err != nil {
+		return err
+	}
+	err = lock.Download(dir, tags, notags, perm, status)
 	if err != nil {
 		return err
 	}
--- a/internal/lock.go
+++ b/internal/lock.go
@@ -89,7 +89,7 @@ func strToFileMode(perm string) (os.FileMode, error) {
 
 // Download gets all the resources in this lock file and moves them to
 // the destination directory.
-func (l *Lock) Download(dir string, tags []string, notags []string, perm string) error {
+func (l *Lock) Download(dir string, tags []string, notags []string, perm string, status bool) error {
 	if stat, err := os.Stat(dir); err != nil || !stat.IsDir() {
 		return fmt.Errorf("'%s' is not a directory", dir)
 	}
@@ -150,11 +150,28 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
 		return fmt.Errorf("nothing to download")
 	}
 	errorCh := make(chan error, total)
-	for _, r := range filteredResources {
+
+	var statusLine *StatusLine
+	if status {
+		statusLine = NewStatusLine(ctx, &filteredResources)
+		err := statusLine.InitResourcesSizes()
+		if err == nil {
+			statusLine.Start(true)
+		} else {
+			statusLine = nil // Do not update or display SL.
+		}
+	}
+
+	for i, r := range filteredResources {
 		resource := r
 		go func() {
+
 			err := resource.Download(dir, mode, ctx)
 			errorCh <- err
+
+			if statusLine != nil {
+				statusLine.Increment(i)
+			}
 		}()
 	}
 	done := 0
--- a/internal/lock_test.go
+++ b/internal/lock_test.go
@@ -115,7 +115,7 @@ func TestDownload(t *testing.T) {
 	lock, err := NewLock(path, false)
 	assert.Nil(t, err)
 	dir := test.TmpDir(t)
-	err = lock.Download(dir, []string{}, []string{}, perm)
+	err = lock.Download(dir, []string{}, []string{}, perm, false)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/internal/status.go
+++ b/internal/status.go
@@ -9,6 +9,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/carlmjohnson/requests"
 	"github.com/dustin/go-humanize"
 	"github.com/rs/zerolog/log"
 )
@@ -98,14 +99,24 @@ func (st *StatusLine) InitResourcesSizes() error {
 	st.totalBytes = 0
 	for i, r := range *st.resources {
 		resource := r
-		httpClient := &http.Client{Timeout: time.Duration(timeoutMs) * time.Millisecond}
-		resp, err := httpClient.Head(resource.Urls[0])
+		headers := http.Header{}
+		err := requests.
+			URL(resource.Urls[0]).
+			Head().
+			CopyHeaders(headers).
+			CheckStatus(http.StatusOK).
+			Fetch(context.Background())
 		if err != nil {
 			log.Debug().Msg("Error fetching resource sizes")
 			return err
 		}
-		st.totalBytes += resp.ContentLength
-		st.resourceSizes[i] = resp.ContentLength
+		ContentLength, err := strconv.Atoi(headers.Get("Content-Length"))
+		if err != nil {
+			log.Debug().Msg("Error fetching resource sizes")
+			return err
+		}
+		st.totalBytes += int64(ContentLength)
+		st.resourceSizes[i] = int64(ContentLength)
 	}
 
 	return nil
--- a/internal/status.go
+++ b/internal/status.go
@@ -1,0 +1,154 @@
+package internal
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strconv"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/dustin/go-humanize"
+	"github.com/rs/zerolog/log"
+)
+
+type StatusLine struct {
+	resources              *[]Resource
+	resourceSizes          []int64
+	totalBytes             int64
+	numBytesDownloaded     int64
+	numResourcesDownloaded int
+	spinI                  int
+	isRunning              atomic.Bool
+	startTime              time.Time
+	sizingErr              error
+	ctx                    context.Context
+	mtx                    sync.RWMutex
+}
+
+const spinChars = "-\\|/"
+const tickMs = 50
+const timeoutMs = 1000
+
+// NewStatusLine creates and initializes a new StatusLine.
+func NewStatusLine(ctx context.Context, resources *[]Resource) *StatusLine {
+	st := StatusLine{}
+	st.resources = resources
+	st.ctx = ctx
+	return &st
+}
+
+// Increment informs the StatusLine that a resource (at index i in resource list) has finished downloading.
+// The SL is printed and Stopped if all resources are downloaded.
+func (st *StatusLine) Increment(i int) {
+	st.mtx.Lock()
+	st.numBytesDownloaded += st.resourceSizes[i]
+	st.numResourcesDownloaded++
+	st.mtx.Unlock()
+
+	fmt.Print(st.GetStatusString() + " ")
+	if st.numResourcesDownloaded == len(*st.resources) {
+		fmt.Println()
+		st.Stop()
+		return
+	}
+}
+
+// Start begins the goroutine and loop that will update/print the status line.
+// Pass true to force SL to update (spinner and second counter) every 50ms.
+func (st *StatusLine) Start(doTick bool) {
+	st.isRunning.Store(true)
+	st.startTime = time.Now()
+	go func() {
+		fmt.Print(st.GetStatusString())
+		for {
+			select {
+			case <-st.ctx.Done():
+				st.Stop()
+				return
+			case <-time.After(tickMs * time.Millisecond):
+				if !doTick {
+					continue
+				}
+			}
+
+			st.mtx.Lock()
+			st.spinI = (st.spinI + 1) % len(spinChars)
+			st.mtx.Unlock()
+
+			fmt.Print(st.GetStatusString() + " ")
+		}
+	}()
+
+}
+
+func (st *StatusLine) Stop() {
+	st.isRunning.Store(false)
+}
+
+// initResourceSizes fetches the size, in bytes, of each resource.
+func (st *StatusLine) InitResourcesSizes() error {
+	log.Debug().Msg("Fetching resource sizes")
+	st.resourceSizes = make([]int64, len(*st.resources))
+	for i := 0; i < len(st.resourceSizes); i++ {
+		st.resourceSizes[i] = 0
+	}
+
+	st.totalBytes = 0
+	for i, r := range *st.resources {
+		resource := r
+		httpClient := &http.Client{Timeout: time.Duration(timeoutMs) * time.Millisecond}
+		resp, err := httpClient.Head(resource.Urls[0])
+		if err != nil {
+			log.Debug().Msg("Error fetching resource sizes")
+			return err
+		}
+		st.totalBytes += resp.ContentLength
+		st.resourceSizes[i] = resp.ContentLength
+	}
+
+	return nil
+}
+
+// GetStatusString composes and returns the status line string for printing.
+func (st *StatusLine) GetStatusString() string {
+	st.mtx.RLock()
+	defer st.mtx.RUnlock()
+
+	var spinner string
+	if st.numResourcesDownloaded < len(*st.resources) {
+		spinner = string(spinChars[st.spinI])
+	} else {
+		spinner = "✔"
+	}
+
+	barStr := "["
+	if st.sizingErr == nil && st.totalBytes > 0 {
+		barLength := 20
+		if st.totalBytes < 20 {
+			barLength = int(st.totalBytes)
+		}
+		squareSize := st.totalBytes / int64(barLength)
+		for i := 0; i < int(st.numBytesDownloaded/squareSize); i += 1 {
+			barStr += "█"
+		}
+		for i := int(st.numBytesDownloaded / squareSize); i < barLength; i += 1 {
+			barStr += " "
+		}
+	}
+	barStr += "]"
+
+	completeStr := strconv.Itoa(st.numResourcesDownloaded) + "/" + strconv.Itoa(len(*st.resources)) + " Resources"
+	byteStr := humanize.Bytes(uint64(st.numBytesDownloaded)) + " / " + humanize.Bytes(uint64(st.totalBytes))
+	elapsedStr := strconv.Itoa(int(time.Since(st.startTime).Round(time.Second).Seconds())) + "s elapsed"
+
+	pad := "          "
+	var line string
+	if st.sizingErr == nil {
+		line = "\r" + spinner + barStr + pad + completeStr + pad + byteStr + pad + elapsedStr // "\r" lets us clear the line.
+	} else {
+		line = "\r" + spinner + "[]" + pad + completeStr + pad + elapsedStr
+	}
+	return line
+}
--- a/internal/status_test.go
+++ b/internal/status_test.go
@@ -1,0 +1,111 @@
+package internal
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/cisco-open/grabit/test"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSpinner(t *testing.T) {
+	resources := createResources(1, t)
+	ctx, _ := context.WithCancel(context.Background())
+	sl := NewStatusLine(ctx, &resources)
+	err := sl.InitResourcesSizes()
+	assert.Nil(t, err)
+
+	sl.Start(true)
+	assert.Equal(t, "\r-[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+	time.Sleep(60 * time.Millisecond) // Give extra ms to allow SL to update.
+	assert.Equal(t, "\r\\[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+	time.Sleep(60 * time.Millisecond)
+	assert.Equal(t, "\r|[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+	time.Sleep(60 * time.Millisecond)
+	assert.Equal(t, "\r/[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+	time.Sleep(60 * time.Millisecond)
+	assert.Equal(t, "\r-[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+}
+
+func TestTimer(t *testing.T) {
+	resources := createResources(1, t)
+	ctx, _ := context.WithCancel(context.Background())
+	sl := NewStatusLine(ctx, &resources)
+	err := sl.InitResourcesSizes()
+	assert.Nil(t, err)
+
+	sl.Start(false)
+	assert.Equal(t, "\r-[      ]          0/1 Resources          0 B / 6 B          0s elapsed", sl.GetStatusString())
+	time.Sleep(1000 * time.Millisecond)
+	assert.Equal(t, "\r-[      ]          0/1 Resources          0 B / 6 B          1s elapsed", sl.GetStatusString())
+	time.Sleep(1000 * time.Millisecond)
+	assert.Equal(t, "\r-[      ]          0/1 Resources          0 B / 6 B          2s elapsed", sl.GetStatusString())
+}
+
+func TestCountersWith2Resources(t *testing.T) {
+	resources := createResources(2, t)
+	ctx, _ := context.WithCancel(context.Background())
+	sl := NewStatusLine(ctx, &resources)
+	err := sl.InitResourcesSizes()
+	assert.Nil(t, err)
+
+	sl.Start(false)
+	assert.Equal(t, "\r-[            ]          0/2 Resources          0 B / 12 B          0s elapsed", sl.GetStatusString())
+	sl.Increment(0)
+	assert.Equal(t, "\r-[██████      ]          1/2 Resources          6 B / 12 B          0s elapsed", sl.GetStatusString())
+	sl.Increment(1)
+	assert.Equal(t, "\r✔[████████████]          2/2 Resources          12 B / 12 B          0s elapsed", sl.GetStatusString())
+
+}
+
+func TestCountersWith1000Resources(t *testing.T) {
+	resources := createResources(1000, t)
+	ctx, _ := context.WithCancel(context.Background())
+	sl := NewStatusLine(ctx, &resources)
+	err := sl.InitResourcesSizes()
+	assert.Nil(t, err)
+
+	sl.Start(false)
+	assert.Equal(t, "\r-[                    ]          0/1000 Resources          0 B / 6.0 kB          0s elapsed", sl.GetStatusString())
+	sl.Increment(0)
+	assert.Equal(t, "\r-[                    ]          1/1000 Resources          6 B / 6.0 kB          0s elapsed", sl.GetStatusString())
+	sl.Increment(1)
+	assert.Equal(t, "\r-[                    ]          2/1000 Resources          12 B / 6.0 kB          0s elapsed", sl.GetStatusString())
+	sl.Increment(2)
+	assert.Equal(t, "\r-[                    ]          3/1000 Resources          18 B / 6.0 kB          0s elapsed", sl.GetStatusString())
+	sl.Increment(3)
+	assert.Equal(t, "\r-[                    ]          4/1000 Resources          24 B / 6.0 kB          0s elapsed", sl.GetStatusString())
+	for i := 4; i < 1000; i++ {
+		sl.Increment(i)
+	}
+	assert.Equal(t, sl.GetStatusString(), "\r✔[████████████████████]          1000/1000 Resources          6.0 kB / 6.0 kB          0s elapsed")
+}
+
+func TestWithCancelledContext(t *testing.T) {
+	resources := createResources(2, t)
+	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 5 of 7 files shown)


internal/ansi_color.go
```diff
@@ -0,0 +1,35 @@
+package internal
```
rabadin, line 1: Usually, CLI tools use colors to denote something (i.e. red for failure, green for success). Since this is a progress I suggest not using colors here. It will simplify things and also make using the progress bar usable in outputs that have a tendency to be messed up by escape sequence.

internal/lock.go
```diff
@@ -86,9 +88,18 @@ func strToFileMode(perm string) (os.FileMode, error) {
 	return os.FileMode(parsed), nil
 }
 
+// Adds commas to number string at hundreds place, thousands place, etc.
```
rabadin, line 91: Maybe you could use https://github.com/dustin/go-humanize here. It would have the advantage of keeping the size of the output concise.

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
 	errorCh := make(chan error, total)
-	for _, r := range filteredResources {
+
+	byteCh := make(chan int64)
+	resourceSizes := make([]int64, 0, len(filteredResources))
+
+	if bar {
+		startTime := time.Now()
+
+		//Loop through resources, fetching their metadata and totalling up their sizes in bytes.
```
rabadin, line 170: nit: it's really a detail but please follow the standard golang comment format and add a space after `//`.

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
 		return fmt.Errorf("nothing to download")
 	}
 	errorCh := make(chan error, total)
-	for _, r := range filteredResources {
+
+	byteCh := make(chan int64)
+	resourceSizes := make([]int64, 0, len(filteredResources))
+
+	if bar {
+		startTime := time.Now()
```
rabadin, line 168: This whole logic belongs in its own file. I suggest refactoring this as a struct so that it is well encapsulated, easier to test and to use.

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+
+		var totalBytes int64 = 0
+		for _, r := range filteredResources {
+			resource := r
+			httpClient := &http.Client{Timeout: 10 * time.Second}
+			resp, err := httpClient.Head(resource.Urls[0])
+			if err != nil {
+				//errorCh <- err	We don't want a failed fetch to crash the whole program -- it is not crucial to know the total download size.
+				//Instead, the print goroutine will check length of resourceSizes and notify the user by printing an error message in place of the byte counter.
+				break
```
rabadin, line 183: Since the size calculation is going to be wrong in case of errors, you probably want to not display the size at all in this case.

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+			}
+			totalBytes += resp.ContentLength
+			resourceSizes = append(resourceSizes, resp.ContentLength)
+		}
+
+		spinChars := [5]string{"-", "\\", "|", "/", "-"}
```
rabadin, line 189: Things like this should be file-level constants. The main advantage is that the compiler will perform additional optimizations on these.

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+		spinI := 0 //Current char in spinChars.
+
+		//SPINNER GOROUTINE.
+		ticker := time.NewTicker(60 * time.Millisecond)
+		go func() {
```
rabadin, line 194: Another, maybe cleaner way to deal with this is to remove the ticker altogether and add the "default" at read time.
e.g.
```
select {
    case b := <-byteCh:
        return b
    case <-time.After(time.Second):
        return -1
    }
```

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+				//Line is yellow while downloading, green when complete.
+				var color string
+				if resourcesDownloaded < len(filteredResources) {
+					color = "yellow"
+				} else {
+					color = "green"
+				}
+
+				//Build progress bar string.
+				barStr := "║"
```
rabadin, line 237: It might be me but these symbol is a tad confusing. How about `[` and `]` for start/end. And an empty space instead of the strange dotted square?

internal/lock.go
```diff
@@ -149,11 +160,131 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+				}
+				barStr += "║"
+
+				completeStr := strconv.Itoa(resourcesDownloaded) + "/" + strconv.Itoa(len(filteredResources)) + " Complete"
+
+				var byteStr string
+				if len(resourceSizes) == len(filteredResources) {
+					byteStr = AddCommas(strconv.Itoa(int(bytesDownloaded))) + "B / " + byteStr + AddCommas(strconv.Itoa(int(totalBytes))) + "B"
+				} else {
+					byteStr = "<ERROR FETCHING RESOURCE SIZES>"
```
rabadin, line 257: Same remark as above: in case of error I suggest simply not printing the sizes.

rabadin (changes requested):
See code related comments inline. In terms of high-level process and design here are my suggestions:
- First, of course, you need to solve the conflicts with the main branch. Do a rebase and solve the conflicts.
- The main design problem that I see with this is that isn't not encapsulated in a way that will let you test this or even use it without embedding the entire code inline like you've done here. I suggest you write this as a struct with methods on it in a separate file. i.e. put the code in a progress.go file and create a Progress stuct that would be instanciated from a list of resources and with methods to update it and get the string to print out ot it.

internal/status.go
```diff
@@ -0,0 +1,131 @@
+package internal
+
+import (
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+type Status_Line struct {
```
rabadin, line 10: Golang generally uses CamelCase for variable and class names. Please follow this convention. This means naming this `StatusLine`.

internal/lock.go
```diff
@@ -149,11 +149,21 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+	}
+
+	for i, r := range filteredResources {
 		resource := r
 		go func() {
+
 			err := resource.Download(dir, mode, ctx)
 			errorCh <- err
+
+			status_line.indexCh <- i
```
rabadin, line 166: Please encapsulate this into a method (`statusLine.increment()`). This way the call sites don't have to manipulate internal fields.

internal/lock.go
```diff
@@ -149,11 +149,21 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
 		return fmt.Errorf("nothing to download")
 	}
 	errorCh := make(chan error, total)
-	for _, r := range filteredResources {
+
+	status_line := Status_Line{filteredResources, make(chan int)}
```
rabadin, line 153: You should have a method named `NewStatusLine(filteredResources)` that builds the status line. This way the call sites don't need to know about the internals of the status line.

rabadin (changes requested):

internal/ansi_color.go
MDylanWilliams: That makes perfect sense. It was good to practice using ASNI codes and adding a new file to the repo.
I'm currently working on the rest of the suggestions. Thank you for your reviewing everything -- like you said, having your code analyzed is a great opportunity.

internal/lock.go
MDylanWilliams: Should the values be passed to this NewStatusLine constructor, then passed to run(), then passed to composeLine()? Or should NewStatusLine store them as fields of the StatusLine struct and run() composeLine() can just access them directly from the pointer/object?

internal/status.go
```diff
@@ -0,0 +1,118 @@
...
+				fmt.Println()
+				break
+			}
+
+		}
+	}()
+
+}
+
+func getResourcesSizes(st *StatusLine, timeoutMs int) ([]int64, int64, bool) {
```
rabadin, line 59: This should be a method on the struct like `run`, this way the caller just have to call object methods.

rabadin:
Yes, anything you need to get the object functional should be stored as a private field.

From: Dylan Williams ***@***.***>
Date: Wednesday, 18 September 2024 at 17:02
To: cisco-open/grabit ***@***.***>
Cc: Raphael Badin (rabadin) ***@***.***>, Comment ***@***.***>
Subject: Re: [cisco-open/grabit] Progress bar (status line) (PR #24)

@MDylanWilliams commented on this pull request.

________________________________

In internal/lock.go<https://github.com/cisco-open/grabit/pull/24#discussion_r1765232826>:

> @@ -149,11 +149,21 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)

               return fmt.Errorf("nothing to download")

        }

        errorCh := make(chan error, total)

-       for _, r := range filteredResources {

+

+       status_line := Status_Line{filteredResources, make(chan int)}

Should the values be passed to this NewStatusLine constructor, then passed to run(), then passed to composeLine()? Or should NewStatusLine store them as fields of the StatusLine struct and run() composeLine() can just access them directly from the pointer/object?

internal/status.go
rabadin, line 59: Ideally this should be a class method.

internal/lock.go
MDylanWilliams: Would it be better for the variables/data to be passed from one function to another (NewStatusLine to run to composeLine) or for them to be stored in StatusLine struct fields that each method can directly access via pointer/object?

rabadin: If you're talking about init data that will be used for the entire life of the object, they should be stored as field, yes.

rabadin:
@MDylanWilliams Please solve the conflicts by rebasing on current state of main.

cmd/download.go
```diff
@@ -8,14 +8,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func init() {
```
rabadin, line 11: This is invalid syntax. Looks like the merge got messed up.
Note: instead of merging you should have *rebased* onto the latest version of main, this would have created a much better history.
But that's okay, we will fix this once we squash this into just one commit.

internal/lock.go
```diff
@@ -150,11 +150,22 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+	}
+
+	for i, r := range filteredResources {
 		resource := r
 		go func() {
+
 			err := resource.Download(dir, mode, ctx)
 			errorCh <- err
+
+			if bar {
```
rabadin, line 166: nit: Instead of checking bar, maybe check if statusLine is empty. This is cleaner in the sense that it will let you potentially refactor the code related to `bar` without having to change other things.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+	bytesDownloaded     int64
+	resourcesDownloaded int
+	spinI               int
+	indexCh             chan int
+	startTime           time.Time
+}
+
+var spinChars = [5]string{"-", "\\", "|", "/", "-"}
+
+// Create, initialize, and start a new StatusLine.
```
rabadin, line 27: See the doc about how inline doc is formatted: https://tip.golang.org/doc/comment#func
It's more idiomatic to start these comments with the name of the function. Like: `NewStatusLine creates, initializes, and starts a new StatusLine`.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+func NewStatusLine(resources []Resource) *StatusLine {
+	st := StatusLine{}
+	st.resources = resources
+	st.indexCh = make(chan int)
+	st.resourceSizes, st.totalBytes, st.sizingSuccess = getResourcesSizes(resources, 10000)
+
```
rabadin, line 33: I suggest removing these empty lines.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+// Begin goroutine and for loop.
+func (st *StatusLine) start() {
+	st.startTime = time.Now()
+
+	go func() {
+		st.spinI = 0
+
+		for {
+
+			//Block until value is inserted into indexCh (>= 0 when resource done downloading, -1 every 50 milliseconds to keep timer and spinner updating).
```
rabadin, line 53: Nit: don't forget the space after `//`

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+				fmt.Println()
+				break
+			}
+
+		}
+	}()
+
+}
+
+func getResourcesSizes(resources []Resource, timeoutMs int) ([]int64, int64, bool) {
```
rabadin, line 87: I suggest refactoring this as a method on StatusLine with the following signature: `func (st *StatusLine) initSizes() error`.
Making it a method you avoid having to pass around things that you end up just setting on the object itself anyway.
I think the timeout should be a module level const, no need to pass it around as a parameter.
It's more idiomatic to return an `error` (see examples in the code) rather than a bool.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+			break
+		}
+		totalBytes += resp.ContentLength
+		resourceSizes[i] = resp.ContentLength
+	}
+
+	return resourceSizes, totalBytes, sizingSuccess
+}
+
+func composeStatusString(bytesDownloaded int64, totalBytes int64, resourcesDownloaded int, numResources int, sizingSuccess bool, spinChars []string, spinI int, startTime time.Time, anyRemaining bool, barLength int) string {
```
rabadin, line 111: Same remark: let's make this a struct method to avoid having to pass a ton of fields that are present on the object:
```
func (st *StatusLine) getStatusString(<what exactly is needed here since all the data is in the fields?>) error {

File: internal/status.go [hunk 21]
```diff
@@ -0,0 +1,147 @@
...
+	}
+	barStr += "]"
+
+	completeStr := strconv.Itoa(resourcesDownloaded) + "/" + strconv.Itoa(numResources) + " Resources"
+
+	var byteStr string
+	if sizingSuccess {
+		byteStr = humanize.Bytes(uint64(bytesDownloaded)) + " / " + humanize.Bytes(uint64(totalBytes))
+	} else {
+		byteStr = "<issue_fetching_resource_sizes>"
```
rabadin, line 138: This should be an error condition and handled by the caller.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+	totalBytes          int64
+	sizingSuccess       bool
+	bytesDownloaded     int64
+	resourcesDownloaded int
+	spinI               int
+	indexCh             chan int
+	startTime           time.Time
+}
+
+var spinChars = [5]string{"-", "\\", "|", "/", "-"}
```
rabadin, line 25: Make this a const to let the compiler optimize this to death.

internal/status_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestComposeStatusString(t *testing.T) {
+	line := composeStatusString(0, 1000, 0, 1, true, []string{"-", "\\", "|", "/", "-"}, 0, time.Now(), true, 20)
+	expected := "\r-[ ]          0/1 Complete          0B / 1,000B          0s Elapsed"
+	assert.Equal(t, expected, line)
+}
```
rabadin, line 14: Let's add the following tests:
- Test the happy, simple path: just 2 small resources, test the initialization of the StatusLine, test that the status string looks okay at first, increment, test again the string, re-increment, test the string, then test the end state of the status string.
- Test the massive case: test the status string when there are 1000 resources in the file
- Test the error cases
  - Can't find the size
  - Passed context gets cancelled
  - other error cases?

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+
+// Create, initialize, and start a new StatusLine.
+func NewStatusLine(resources []Resource) *StatusLine {
```
rabadin, line 28: The method should take a `ctx context.Context`. See https://pkg.go.dev/context for details.

internal/status.go
```diff
@@ -0,0 +1,147 @@
...
+
+	go func() {
+		st.spinI = 0
+
+		for {
+
+			//Block until value is inserted into indexCh (>= 0 when resource done downloading, -1 every 50 milliseconds to keep timer and spinner updating).
+			var i int
+			select {
+			case i = <-st.indexCh:
```
rabadin, line 56: You should add a case for when the passed context (see my other comment about this) is cancelled:
```
		case <-st.ctx.Done():
			return ctx.Err()
```

rabadin (changes requested):
This is starting to look good. Of course I have more comments, see inline.
Two main things:
- You should have a done a rebase instead of a merge. We can fix this later;
- The merge resulted in non-compling code;
- You need to start adding the UT once you've addressed the new batch of comments.

internal/status.go
```diff
@@ -20,15 +21,17 @@ type StatusLine struct {
...
 	startTime              time.Time
 	sizingErr              error
+	ctx                    context.Context
 }
 
 var spinChars = [5]string{"-", "\\", "|", "/", "-"}
 
 // NewStatusLine creates, initializes, and starts a new StatusLine.
-func NewStatusLine(resources []Resource) (*StatusLine, error) {
+func NewStatusLine(resources []Resource, ctx context.Context) (*StatusLine, error) {
```
rabadin, line 30: It's a detail but usually the context is passed as the first argument.

internal/lock.go
MDylanWilliams: In our meeting we discussed modifying the Download function to continue if sizing fails and display a warning to the user. 
Would you like to display a modified status line that only includes the timer, resource ticker, and bar (which could display resources rather than bytes, as it was a while ago)?
Or would you rather something like "Issue getting file sizes" be printed and the status line be omitted completely?

internal/status.go
```diff
@@ -0,0 +1,148 @@
...
+				fmt.Println()
+				return
+			}
+
+		}
+	}()
+
+}
+
+func (st *StatusLine) getResourcesSizes(resources []Resource, timeoutMs int) {
```
rabadin, line 87: This should probably be renamed `initResourcesSizes` now that it doesn't return anything. And it should return an error in case something goes wrong.

internal/status.go
```diff
@@ -0,0 +1,148 @@
...
+	completeStr := strconv.Itoa(st.numResourcesDownloaded) + "/" + strconv.Itoa(len(st.resources)) + " Resources"
+
+	var byteStr string
+	if st.sizingErr == nil {
+		byteStr = humanize.Bytes(uint64(st.numBytesDownloaded)) + " / " + humanize.Bytes(uint64(st.totalBytes))
+	} else {
+		byteStr = "<issue_fetching_resource_sizes>"
+	}
+
+	elapsedStr := strconv.Itoa(int(time.Since(st.startTime).Round(time.Second).Seconds())) + "s Elapsed"
```
rabadin, line 142: s/Elapsed/elapsed/

internal/status.go
MDylanWilliams: Right now, if an error occurs, initResourceSizes stores the error in the sizingErr field. This is then returned from NewStatusLine for the caller to handle and also used in composeStatusString to change the output. 

Should this be changed?

internal/status.go
MDylanWilliams: Do you mean that the variable should be renamed?

rabadin: I mean E -> e

internal/status.go
rabadin: Okay, that's probably fine. But you should still rename the method since it doesn't return anything.

internal/status_test.go
MDylanWilliams: How would you recommend I go about generating 1000 test resources?

rabadin: Something like this:
```
	content := `abcdef`
	port := test.TestHttpHandler(content, t)
	resources := []Resource{}
	algo := "sha256"
	for i := 0; i < 1000; i++ {
		resource := Resource{Urls: []string{fmt.Sprintf("http://localhost:%d/test%d.html", port, i)}, Integrity: fmt.Sprintf("%s-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE=", algo), Tags: []string{}, Filename: ""}
		resources = append(resources, resource)
	}
	assert.Len(t, resources, 1000)
```

internal/status.go
MDylanWilliams: Would it be better for initResourceSizes to return the error (to better match the pattern of functions returning their errors in the rest of the codebase) -- then NewStatusLine could store it in sizingErr (for later use in composeStatusString) and finally return it to the caller?

rabadin: I think it would be preferable, yes. That's the normal pattern that golang developers would expect.

internal/lock.go
rabadin: Up to you. I would opt for the simplest option: not display anything in case of error.

internal/status_test.go
```diff
@@ -1,12 +1,47 @@
...
+
+	// Test StatusLine initialization and initResourcesSizes().
+	st, err := NewStatusLine(ctx, resources)
+	assert.Nil(t, err)
+
+	// Test GetStatusString() and Increment().
+	st.Start(false)
+	assert.Equal(t, "\r-[                    ]          0/1000 Resources          0 B / 6.0 kB          0s elapsed", st.GetStatusString())
+	st.Increment(0)
+	time.Sleep(10 * time.Millisecond) // Give StatusLine loop time to update fields.
```
MDylanWilliams, line 36: This is an arbitrary length of time that worked for me on my machine. Alternatively, TestGetStatusString could create and pass a channel that the SL could use to tell TestGetStatusString it has updated its internal states/fields and is ready to be checked -- Sleep would be replaced with a blocking select-case statement. I would assume this would be much more reliable than these Sleep calls

internal/status.go
```diff
@@ -0,0 +1,152 @@
...
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/dustin/go-humanize"
+)
+
+type StatusLine struct {
+	resources []Resource
```
rabadin, line 14: It's preferable to have a *pointer* to a list of resources, to avoid duplicating data.

internal/status.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+
+		}
+	}()
+
+}
+
+// initResourceSizes fetches the size, in bytes, of each resource in the provided list.
+// An error, if encountered, is stored in sizingErr.
+func (st *StatusLine) initResourcesSizes(resources []Resource) error {
```
rabadin, line 89: No need to pass the list of resources, this method should use `st.resources`.

rabadin (changes requested):

internal/lock.go
```diff
@@ -150,11 +150,26 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
 		return fmt.Errorf("nothing to download")
 	}
 	errorCh := make(chan error, total)
-	for _, r := range filteredResources {
+
+	var statusLine *StatusLine
+	if status {
+		statusLine, err = NewStatusLine(ctx, &filteredResources)
+		if err != nil {
+			return err // Do not continue if resource sizing failed.
```
rabadin, line 158: This would fail the entire download step right? We said we should just print a message and but continue the download in this case.

internal/status.go
rabadin: You still need to make this a `const`

internal/status.go
```diff
@@ -111,6 +114,9 @@ func (st *StatusLine) initResourcesSizes() error {
 
 // GetStatusString composes and returns the status line string for printing.
 func (st *StatusLine) GetStatusString() string {
+	st.mtx.Lock()
```
rabadin, line 117: I don't understand why this is needed... you're only *reading* data here.

rabadin (changes requested):

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+				return
+			}
+
+		}
+	}()
+
+}
+
+// initResourceSizes fetches the size, in bytes, of each resource in the provided list.
+// An error, if encountered, is stored in sizingErr.
```
rabadin, line 91: You mention sizingErr here but that describes how the method it used, not what it does. You should remove this last sentence from the doc.

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+func (st *StatusLine) initResourcesSizes() error {
+	fmt.Print("\rFetching resource sizes...")
```
rabadin, line 93: This should be either removed or added as a debug comment.

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+		st.resourceSizes[i] = 0
+	}
+
+	st.totalBytes = 0
+	for i, r := range *st.resources {
+		resource := r
+		httpClient := &http.Client{Timeout: time.Duration(timeoutMs) * time.Millisecond}
+		resp, err := httpClient.Head(resource.Urls[0])
+		if err != nil {
+			fmt.Println("\rError fetching resource sizes")
```
rabadin, line 105: No print statements please, use the logger.

internal/lock_test.go
```diff
@@ -115,7 +115,7 @@ func TestDownload(t *testing.T) {
 	lock, err := NewLock(path, false)
 	assert.Nil(t, err)
 	dir := test.TmpDir(t)
-	err = lock.Download(dir, []string{}, []string{}, perm)
+	err = lock.Download(dir, []string{}, []string{}, perm, true)
```
rabadin, line 118: This last param should be false: here we're testing the lock code. We don't want to exercise the status code.
But we do want to add a full download with status enabled in status_test.go

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+func (st *StatusLine) Start(doTick bool) {
+	st.startTime = time.Now()
+	go func() {
+		fmt.Print(st.GetStatusString())
+		for {
+			// Block until value is inserted into indexCh (>= 0 when resource finishes downloading, -1 every 50 milliseconds to keep timer and spinner updating).
+			var i int
+			select {
+			case i = <-st.indexCh:
+			case <-time.After(50 * time.Millisecond):
```
rabadin, line 57: I think you want to make this duration a const. This way you can refer to it in the tests if needed.

internal/status_test.go
```diff
@@ -0,0 +1,47 @@
...
+	resources := []Resource{}
+	algo := "sha256"
+	for i := 0; i < 1000; i++ {
+		resource := Resource{Urls: []string{fmt.Sprintf("http://localhost:%d/test%d.html", port, i)}, Integrity: fmt.Sprintf("%s-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE=", algo), Tags: []string{}, Filename: ""}
+		resources = append(resources, resource)
+	}
+	ctx, _ := context.WithCancel(context.Background())
+
+	// Test resource generation.
+	assert.Len(t, resources, 1000)
```
rabadin, line 26: You can remove this. I just added it when I gave you this code to show what it was doing, but there is no need to test our own pre-test code.

MDylanWilliams:
How would you recommend I go about testing context cancellation? The SL could close a channel when terminated, or it could set a new internal field (e.g. a bool called "isRunning") to false. Or another way?

rabadin:
> How would you recommend I go about testing context cancellation? The SL could close a channel when terminated, or it could set a new internal field (e.g. a bool called "isRunning") to false. Or another way?

That sounds reasonable. I think you should also remove the parameter to `InitResourcesSizes`; I don't think it's useful. https://gist.github.com/rabadin/44a3d2a1f0c91e3f9e7c2b76440f3989

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+		spinner = "✔"
+	}
+
+	barStr := "["
+	if st.sizingErr == nil {
+		barLength := 20
+		if st.totalBytes < 20 {
+			barLength = int(st.totalBytes)
+		}
+		squareSize := st.totalBytes / int64(barLength)
```
rabadin, line 133: This can break if `barLength` is zero. You need to deal gracefully with this case.

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+		for i := 0; i < int(st.numBytesDownloaded/squareSize); i += 1 {
```
rabadin, line 134: Can squareSize  be zero here? If yes you need to deal with this.

internal/status.go
```diff
@@ -0,0 +1,158 @@
...
+			barStr += "█"
+		}
+		if st.numResourcesDownloaded < len(*st.resources) {
+			barStr += " "
+		}
+		for i := int(st.numBytesDownloaded/squareSize) + 1; i < barLength; i += 1 {
```
rabadin, line 140: Ditto

rabadin (changes requested):

MDylanWilliams:
> That sounds reasonable. I think you should also remove the parameter to `InitResourcesSizes`; I don't think it's useful. https://gist.github.com/rabadin/44a3d2a1f0c91e3f9e7c2b76440f3989

Yes, the timeOut parameter does just exist for testing. Other than a parameter, how can we force the size requests to fail to test that case?

MDylanWilliams:
The if statement now reads `if st.sizingErr == nil && st.totalBytes > 0 {` in case totalBytes is still somehow 0 even after sizing.

barLength is set to 20 or set to totalBytes, and totalBytes is > 0, so barLength is always > 0 and st.totalBytes / barLength cannot result in a divide-by-zero error. 

Please let me know if I've completely overlooked something.

rabadin:
@MDylanWilliams  Also see the test failure: https://github.com/cisco-open/grabit/actions/runs/11699941753/job/32583452591?pr=24

MDylanWilliams:
Moving the SL field updates from the for loop to Increment would allow us to omit the Sleep calls, right?

rabadin:
@MDylanWilliams As discussed during the meeting, please squash all your commits into one, carefully editing the commit message (see Jeff's suggestions) and then I think we can land this.

MDylanWilliams:
I noticed that an extra 'd' will get printed at the end of the SL ("5s elapsedd").

I believe this is because "/r" is used to move the cursor back to the beginning of the terminal's current line, and if the newly printed line is shorter than the last/existing line, the remaining characters are left uncleared. 

Printing "\033[2K" to clear the current terminal line works and looks fine on my mac, but it causes flashing on my Windows laptop.

MDylanWilliams:
Also, despite my thinking it was solved, I am still occasionally getting the bug where the SL fails to do its final update.

jsquyres:
@MDylanWilliams Can you squash down to a single commit?

MDylanWilliams:
Let me know if the squashed commit message needs modified or shortened.

rabadin:
> @MDylanWilliams Somehow you ended up with a bunch of conflicts. ☹️

Not only this... but the diff includes way more than what was in this change...

rabadin (approved):

PR #25, opened 2024-10-02
Optimization Portion Pull Requestion 

777Denoiser:
## Description

Finished implementing the opti-downloader into the grabit codebase, i already crafted the downloader.go into the code and tested it in a secure environment to test for Invalid Hashes, Validation of Downloads, and Tested for Invalid URLs to handle redudant downloads, refer to the comments in the downloader.go files for more information. In the main.go I added downloader in the newrootcmd, edited the download.go, main.go, lock.go, root.go to implement the downloader after testing it in a local tester. 

this change will implement a new downloader into the grabit code base that will do the following: 
// this implementation includes a feature to skip unnecessary downloads if the file already exists.
// DownloadFile function checks for existing files and verifies their hash, and only downloads if necessary.
// This also includes error handling for hash mismatches ensuring data integrity improving efficiency
// and avoiding redundant downloads while maintaining file accuracy.

Everything is functional with the new downloader; the problem is the dynamic assets have changing sha hashes therefore they fail automatically which is a bug fix i need to implement otherwise the downloader works with static assets that have sha hashes that stay static.

Please provide a meaningful description of what this change will do, or is for. Bonus points for including links to
related issues, other PRs, or technical references.

Note that by _not_ including a description, you are asking reviewers to do extra work to understand the context of this
change, which may lead to your PR taking much longer to review, or result in it not being reviewed at all.

## Type of Change

- [ ] Bug Fix
- [Y] New Feature
- [ ] Breaking Change
- [Y] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [Y] I have read the [contributing guidelines](/CONTRIBUTING.md)
- [Y] Existing issues have been referenced (where applicable)
- [Y] I have verified this change is not present in other open pull requests
- [Y] Functionality is documented
- [Y] All code style checks pass
- [Y] New code contribution is covered by automated tests
- [Y] All new and existing tests pass


.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
rabadin, line 1: You've probably committed the content of .idea before adding this line... because the content of .idea is committed. You definitely want to remove it.

rabadin (changes requested):
That's a good start but you've re-implemented a lot of stuff that exists already. I don't think you need to add more packages. You should be able to just refine the ones that exists. Also: there should be no need to test the command layer. This change should be done completely "under the covers".

rabadin:
@777Denoiser  Please do not close this PR, especially if you're going to create another: continue iterating on this request so that we have the full history.

jsquyres:
> @rabadin sorry I had to create another PR which was #26 with my updated add-ons and future add-ons, is there a way to preserve the closed PR? I hope you can still access the history of the #25 PR. apologies I won't be closing any future PRs. thank you!

It is better to just keep pushing to your original branch (i.e., the original PR).  The closed PR is preserved, but it's separate from your new PR.

PR #28, opened 2024-10-28
Optimization

777Denoiser:


jsquyres (changes requested):
It is unexpected that you closed #26 and then opened this PR (from the same source branch).

I think nearly all of my comments on #26 still apply to this PR.

jsquyres (changes requested):
Overall, I think the sentiment is the same here: some of these are interesting changes, but we really want to see a PR that is *only* the "don't re-download if the file already exists in the filesystem and validates properly".

Please move the other improvements / fixes to other PRs where they can be discussed separately.

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
+package internal
```
rabadin, line 1: Can you please move this test to `download_test.go`. This is the place where we test all high-level download use-cases.

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+func TestDownloadOptimization(t *testing.T) {
+	// Setup test server
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("test content"))
+	}))
+	defer ts.Close()
+
+	tmpDir := t.TempDir()
```
rabadin, line 21: Instead of doing this manually, you should utils.go:TmpFile.

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+		err = lock.Download(tmpDir, nil, nil, "")
+		require.NoError(t, err)
+	})
+
+	t.Run("Invalid File Redownloaded", func(t *testing.T) {
+		lockPath := filepath.Join(lockDir, "grabit.lock")
+		lock, err := NewLock(lockPath, true)
+		require.NoError(t, err)
+
+		err = lock.AddResource([]string{testUrl}, RecommendedAlgo, nil, "test.txt")
```
rabadin, line 49: I see that all your tests use the same file... isn't this a bit dangerous in terms of stability/correctness?

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+	err := os.MkdirAll(lockDir, 0755)
+	require.NoError(t, err)
+
+	testUrl := ts.URL + "/test.txt"
+	testContent := []byte("test content")
+	testFile := filepath.Join(tmpDir, "test.txt")
+	err = os.WriteFile(testFile, testContent, 0644)
+	require.NoError(t, err)
+
+	t.Run("Valid File Not Redownloaded", func(t *testing.T) {
```
rabadin, line 32: Nice touch using `t.Run`! Could we go one step further and use `t.Parallel()`?

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+		require.NoError(t, err)
+
+		err = os.WriteFile(testFile, []byte("corrupted"), 0644)
+		require.NoError(t, err)
+
+		err = lock.Download(tmpDir, nil, nil, "")
+		require.NoError(t, err)
+	})
+
+	t.Run("Multiple Resources", func(t *testing.T) {
```
rabadin, line 59: What is this testing? Isn't this already completely tested elsewhere?

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+		for _, url := range urls {
+			err = lock.AddResource([]string{url}, RecommendedAlgo, nil, "")
+			require.NoError(t, err)
+		}
+
+		err = lock.Download(tmpDir, nil, nil, "")
+		require.NoError(t, err)
+	})
+
+	t.Run("Resource Management", func(t *testing.T) {
```
rabadin, line 78: What is this testing? Isn't this already completely tested elsewhere?

internal/optimization_test.go
```diff
@@ -0,0 +1,90 @@
...
+		require.NoError(t, err)
+
+		err = lock.AddResource([]string{testUrl}, RecommendedAlgo, nil, "test.txt")
+		require.NoError(t, err)
+
+		err = lock.Download(tmpDir, nil, nil, "")
+		require.NoError(t, err)
+	})
+
+	t.Run("Invalid File Redownloaded", func(t *testing.T) {
```
rabadin, line 44: I don't think we should blindly overwrite a file if it doesn't have the expected content. It's generally a bad idea to destroy content. In this case, grabit should error.

internal/resource.go
```diff
@@ -90,25 +95,38 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
...
+			ok = true
+			continue
+		}
+
+		// Download file in the target directory
 		lpath, err := GetUrlToDir(u, dir, ctx)
 		if err != nil {
 			downloadError = err
-			break
+			continue
```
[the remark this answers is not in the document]
rabadin: I think this is totally correct and it's fixing a bug. That being said, I think we should fix the bug *and* add a UT for it. I'd rather do it in a separate PR. This is what I'm doing in https://github.com/cisco-open/grabit/pull/29. Please have a look.

rabadin (changes requested):

internal/optimization_test.go
jsquyres: > I don't think we should blindly overwrite a file if it doesn't have the expected content. It's generally a bad idea to destroy content. In this case, grabit should error.

Good clarification.  I think @777Denoiser did this because that's what I advised to do (if it doesn't validate, re-download it).  If we want grabit to error instead, cool.

Question: should grabit error on the first problem it finds?  Or should it still potentially download all the resources that it can, and list all errors that it found at the end of its run?

internal/resource.go
```diff
@@ -90,6 +106,9 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 	}
 	var downloadError error = nil
 	for _, u := range l.Urls {
+main
```
rabadin, line 109: Looks like a leftover.

internal/resource.go
```diff
@@ -102,34 +121,69 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 			return err
 		}
 
+main
```
rabadin, line 124: Looks like a leftover.

cmd/dowload_test.go
```diff
@@ -141,4 +198,5 @@ func TestRunDownloadTriesAllUrls(t *testing.T) {
 	for _, file := range []string{"test.html"} {
 		test.AssertFileContains(t, fmt.Sprintf("%s/%s", outputDir, file), content)
 	}
+main
```
rabadin, line 201: Looks like a leftover.

rabadin (changes requested):

cmd/dowload_test.go
```diff
@@ -124,6 +130,57 @@ func TestRunDownloadFailsIntegrityTest(t *testing.T) {
...
+		require.NoError(t, err)
+
+		err = lock.AddResource([]string{testUrl}, internal.RecommendedAlgo, nil, filepath.Base(testFile))
+		require.NoError(t, err)
+
+		err = os.WriteFile(testFile, []byte("corrupted"), 0644)
+		require.NoError(t, err)
+
+		err = lock.Download(tmpDir, nil, nil, "")
+		require.NoError(t, err)
```
rabadin, line 181: Didn't we agree that in this case the code should error? That grabit should refuse to overwrite an incorrect file...?

rabadin (changes requested):

internal/resource.go
```diff
@@ -109,18 +130,51 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
...
-			err = os.Chmod(resPath, mode.Perm())
-			if err != nil {
+			if err := os.Chmod(resPath, mode.Perm()); err != nil {
 				return err
 			}
 		}
 		ok = true
 	}
+
+	if !ok && downloadError != nil {
```
rabadin, line 175: This looks like a repeat of the code below...?

internal/resource.go
```diff
@@ -63,13 +63,32 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
+// Modify the GetUrlToDir function to ensure proper file cleanup
+func GetUrlToDir(u string, targetDir string, ctx context.Context) 
+(string, error) {
 	// create temporary name in the target directory.
 	h := sha256.New()
 	h.Write([]byte(u))
 	fileName := filepath.Join(targetDir, fmt.Sprintf(".%s", hex.EncodeToString(h.Sum(nil))))
-	return getUrl(u, fileName, ctx)
+
+	// Remove existing file if present
```
rabadin, line 75: We're using a random filename so the file is sort of guaranteed not to be present... so I think we don't need this.

internal/resource.go
```diff
@@ -63,13 +63,32 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
+	os.Remove(fileName)
+
+	// Create new file with immediate close
```
rabadin, line 78: Why the need to do this?

internal/resource.go
```diff
@@ -63,13 +63,32 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
+	file, err := os.Create(fileName)
+	if err != nil {
+		return "", fmt.Errorf("failed to create file: %v", err)
+	}
+	file.Close()
+
+	fileName, err = getUrl(u, fileName, ctx)
+	if err != nil {
+		os.Remove(fileName)
```
rabadin, line 87: I agree with the intent here but I don't think this cleanup belongs here. If we really want to do it we should do it in `getUrl`: in the error case there, check if the file exists and if it does, remove it.

internal/resource.go
```diff
@@ -63,13 +63,32 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
 	return fileName, nil
 }
 
-// GetUrlToDir downloads the given resource to the given directory and returns the path to it.
-func GetUrlToDir(u string, targetDir string, ctx context.Context) (string, error) {
+// GetUrlToDir downloads the given resource to a temporary file and returns the path to it.
+// Modify the GetUrlToDir function to ensure proper file cleanup
```
rabadin, line 67: I think you should remove this comment, see my other comment below.

rabadin (changes requested):

internal/lock_test.go
```diff
@@ -28,12 +28,30 @@ func TestNewLockValid(t *testing.T) {
 `)
 	lock, err := NewLock(path, false)
 	assert.Nil(t, err)
+
 	assert.Equal(t, 1, len(lock.conf.Resource))
 	statement := lock.conf.Resource[0]
 	assert.Equal(t, "sha256-asdasdasd", statement.Integrity)
 	assert.Equal(t, []string{"tag1", "tag2"}, statement.Tags)
+
+	runtime.GC() // Release resources
```
rabadin, line 37: Why are you adding this? There is absolutely no need for this.

internal/lock_test.go
```diff
@@ -28,12 +28,30 @@ func TestNewLockValid(t *testing.T) {
...
 }
 
+func TestDuplicateResource(t *testing.T) {
```
rabadin, line 40: Why are you moving this around and slightly modifying it? I don't see why you're doing this or the need for it.

rabadin:
> I am not sure how to proceed further to closing this PR i would like some guidance. @rabadin @jsquyres

The error is in red in the CI run output: `Error: cmd/dowload_test.go:12:2: "runtime" imported and not used`.
If you remove this the CI passes.

internal/resource.go
```diff
@@ -109,27 +113,56 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 			localName = path.Base(u)
 		}
 		resPath := filepath.Join(dir, localName)
-		err = os.Rename(lpath, resPath)
+
+		// Check existing file first
+		if _, err := os.Stat(resPath); err == nil {
```
rabadin, line 118: This code is a bit wasteful: you're doing `os.Stat` multiple times. It's not *too* bad of course but you could structure this better to be more efficient *and* more clear for the reader.

cmd/dowload_test.go
```diff
@@ -59,34 +65,10 @@ func TestRunDownloadWithTags(t *testing.T) {
 	cmd.SetArgs([]string{"-f", testfilepath, "download", "--tag", "tag", "--dir", outputDir})
 	err := cmd.Execute()
 	assert.Nil(t, err)
-	for _, file := range []string{"test.html"} {
-		test.AssertFileContains(t, fmt.Sprintf("%s/%s", outputDir, file), content)
-	}
-}
-
-func TestRunDownloadWithoutTags(t *testing.T) {
```
jsquyres, line 67: Is there a reason to delete this test?

rabadin: In the interest of keeping the change minimal you should just keep the existing tests and add *only* the tests that are necessary to test your change, no remove existing existing tests.

jsquyres: I took the liberty of un-resolving this comment: let's go ahead and put the test back.  As @rabadin noted: let's keep all existing tests and only *add* new tests that are testing your change.

internal/resource.go
rabadin: You've marked this resolved but you have not answered my question.

internal/resource.go
rabadin: I'm still not sure about this. Why do we need this code at all?

internal/resource.go
```diff
@@ -90,17 +104,7 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 	}
 	var downloadError error = nil
 	for _, u := range l.Urls {
-		// Download file in the target directory so that the call to
```
rabadin, line 93: This comment needs to stay, it's quite important to explain why we're doing this.

rabadin (changes requested):

internal/resource.go
rabadin: | this is not included and has been deleted in the final 1 commit apologies for not communicating that with you raphael, last week in the last few commits i did it was fixed

Do you mean that this code was removed? I see it's still very much here.

internal/resource.go
rabadin: I don't see the result of this discussion in the code.

internal/resource.go
rabadin: I don't see the comment re-instated.

rabadin (changes requested):

internal/resource.go
rabadin: Code is still here...

rabadin (changes requested):

jsquyres:
> this is what i see on my end for the lines 74 to 78, maybe you are referencing an older resource.go,
> 
> i made sure to remove any instances of os.Remove(fileName).

Right, and what you see is what @rabadin is asking about: you deleted this line:

```go
return getUrl(u, fileName, ctx)
```

and replaced it with a bunch of new code.  There's no reason to do that on this PR.  You can go ahead and remove all that new code from `GetUrlToDir()` and just put that one line back.

jsquyres:
> @jsquyres not a problem at all, thank you for pointing that out, i was confused so thanks! i managed to delete the code i added and returned the line back to where it was before.

Minor problem -- it doesn't compile.  Can you fix?

```
Error: internal/resource.go:71:2: declared and not used: fileName
Error: internal/resource.go:72:20: undefined: filename
```

jsquyres:
Still got some compile errors:

```
Error: cmd/dowload_test.go:150:41: not enough arguments in call to lock.Download
	have (string, nil, nil, string)
	want (string, []string, []string, string, bool)
Error: cmd/dowload_test.go:1[7](https://github.com/cisco-open/grabit/actions/runs/12160436133/job/33912772022?pr=28#step:4:8)6:41: not enough arguments in call to lock.Download
	have (string, nil, nil, string)
	want (string, []string, []string, string, bool)
```

jsquyres:
I think the confusing point here is:

* This PR compiles and runs the tests ok.
* **BUT** CI is failing.

Why?

The reason why is because github is effectively making a pretend merge commit from this branch and main, and then doing the compile and running the tests there.  *That's* where the compile fail / test fails are occurring.

My suggestion would be to rebase this PR on top of main.  Then build and run the tests, and you should see the same failures that CI is showing.

rabadin:
> i am not sure but i hope i fixed the problem that was with the PR, even though the CI test passes i know there will be conflicts which is why i am not sure.

Conflicts happen when something lands on the target branch that modifies the exact same lines that your change touches. Unless something lands before your PR does, there won't be conflicts.

rabadin:
@777Denoiser Can you please squash your commits into just one commit?

rabadin (approved):

jsquyres (approved):

PR #30, opened 2024-11-05
Implemented Artifactory upload and download handling

eeasaa01:
# Implemented Artifactory Cache Integration

## Changes
- Added GRABIT_ARTIFACTORY_TOKEN support
- Implemented cache download priority
- Added fallback to source URLs
- Added integrity validation
- Added unit tests for cache functionality
- Added integration tests
- Fixed logging levels
- Added proper error handling

## Testing Done
- Added TestGrabitRequirements for comprehensive testing
- Tested token validation
- Tested cache download and fallback
- Tested NO_CACHE_UPLOAD
- Tested error handling
- All tests passing

## Changes Made Based on Feedback
- Fixed authentication error handling
- Adjusted log levels for cache operations
- Improved error messages
- Added proper validation at each step
- Added comprehensive test coverage

## Checklist
- [x] I have read the contributing guidelines
- [x] Functionality is documented
- [x] All code style checks pass
- [x] New code contribution is covered by automated tests
- [x] All new and existing tests pass


cmd/add.go
```diff
@@ -42,13 +55,75 @@ func runAdd(cmd *cobra.Command, args []string) error {
...
-	err = lock.AddResource(args, algo, tags, filename)
+	cache, err := cmd.Flags().GetString("cache")
 	if err != nil {
 		return err
 	}
+
+	// Check if ARTIFACTORY_TOKEN is set if cache URL is provided
+	token := os.Getenv("ARTIFACTORY_TOKEN")
+	if token == "" && cache != "" {
+		log.Warn().Msg("ARTIFACTORY_TOKEN is not set; cache upload may fail if authentication is required.")
```
jsquyres, line 66: This should be an error, right?  I don't think there will ever be a case where you can upload to Artifactory without authentication.

cmd/add.go
```diff
@@ -42,13 +55,75 @@ func runAdd(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// If it's a local file path and cache URL is provided, upload the file
+	if !isRemote && cache != "" {
+		err := uploadToArtifactory(args[0], cache, token)
+		if err != nil {
+			return fmt.Errorf("failed to upload to Artifactory: %w", err)
+		}
+		log.Info().Msgf("Successfully uploaded resource to Artifactory: %s", cache)
```
jsquyres, line 75: This should probably be debug.  In most successful cases, grabit has no output / is silent.

cmd/download.go
```diff
@@ -47,9 +55,22 @@ func runFetch(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	// Ensure ARTIFACTORY_TOKEN is set if cache URLs require it (new part starts here)
+	token := os.Getenv("ARTIFACTORY_TOKEN")
+	if token == "" {
```
jsquyres, line 61: I don't think that you *always* want to warn if `ARTIFACTORY_TOKEN` is not set -- you only want to *error* if you *need* the Artifactory token and it's not set (i.e., if you have a cache that you need to download from).

cmd/download.go
```diff
@@ -47,9 +55,22 @@ func runFetch(cmd *cobra.Command, args []string) error {
...
+		log.Warn().Msg("ARTIFACTORY_TOKEN is not set; any cache URLs requiring authentication may fail.")
+	} else {
+		log.Debug().Msg("ARTIFACTORY_TOKEN is set; proceeding with potential cache authentication.")
+	}
+	// New part ends here
```
jsquyres, line 66: This is a spurious comment -- go ahead and remove it.

cmd/download.go
```diff
@@ -23,14 +27,18 @@ func addDownload(cmd *cobra.Command) {
...
 	lockFile, err := cmd.Flags().GetString("lock-file")
 	if err != nil {
 		return err
 	}
+
+	// Open lock file as a local file, not a URL
 	lock, err := internal.NewLock(lockFile, false)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to open lock file '%s': %w", lockFile, err)
```
jsquyres, line 39: Is the error message that comes back from `internal.NetLock()` not sufficient?  If not, should it be addressed there, not here?

cmd/download.go
```diff
@@ -47,9 +55,22 @@ func runFetch(cmd *cobra.Command, args []string) error {
...
+	// New part ends here
+
+	// Proceed with the downloading logic
 	err = lock.Download(dir, tags, notags, perm)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to download resources: %w", err)
 	}
+
+	log.Info().Msg("Download completed successfully")
```
jsquyres, line 74: Same comment as mentioned previously: this should likely be Debug.  When grabit succeeds, it is generally silent.

cmd/download.go
```diff
@@ -47,9 +55,22 @@ func runFetch(cmd *cobra.Command, args []string) error {
...
+	} else {
+		log.Debug().Msg("ARTIFACTORY_TOKEN is set; proceeding with potential cache authentication.")
+	}
+	// New part ends here
+
+	// Proceed with the downloading logic
 	err = lock.Download(dir, tags, notags, perm)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to download resources: %w", err)
```
jsquyres, line 71: Same comment as mentioned previously: is the error message from the failed function insufficient?  Should it be addressed there?

internal/hash.go
```diff
@@ -21,38 +34,78 @@ var algos = map[string]Hasher{
 
 var RecommendedAlgo = "sha256"
 
+// Hasher defines a function that returns a hash.Hash.
```
jsquyres, line 37: These extra comments are helpful, but they're not really part of the functionality of this PR.  It would probably be good to remove them from this PR and put them in a separate PR.

internal/hash.go
```diff
@@ -21,38 +34,78 @@ var algos = map[string]Hasher{
...
 
+// init initializes the list of available algorithms.
 func init() {
 	initAlgoList()
 }
 
+// initAlgoList initializes the list of available algorithms and verifies the recommended algorithm.
 func initAlgoList() {
-	algoList := make([]string, 0, len(algos))
+	algoNames := make([]string, 0, len(algos))
```
jsquyres, line 56: This variable name change may be helpful, but it's not really part of this PR.  Make this PR just be about Artifactory.  If you want to propose additional changes, make those be in separate PRs.

internal/lock.go
```diff
@@ -26,178 +28,295 @@ type config struct {
...
 		} else {
 			return nil, fmt.Errorf("file '%s' does not exist", path)
 		}
 	}
+
 	var conf config
 	file, err := os.Open(path)
 	if err != nil {
-		return nil, error
+		return nil, err
```
jsquyres, line 45: This looks like a legit bug fix!

But again: it should be in a separate PR -- not part of this PR.

internal/lock.go
```diff
@@ -26,178 +28,295 @@ type config struct {
...
 	}
+	defer file.Close()
```
jsquyres, line 47: @rabadin Is this a legit bug fix, too?

internal/lock.go
```diff
@@ -26,178 +28,295 @@ type config struct {
...
+
 	d := toml.NewDecoder(file)
-	err = d.Decode(&conf)
-	if err != nil {
-		return nil, err
+	if err := d.Decode(&conf); err != nil {
+		return nil, fmt.Errorf("failed to decode lock file: %w", err)
```
jsquyres, line 51: I'll say this one more time on this PR and then stop making the same comment repeatedly: this is a style change that doesn't really have anything to do with the Artifactory functionality.  If you want to propose this change, put it in a separate PR.  Go ahead and remove this change from this PR.

internal/lock.go
```diff
@@ -6,16 +6,18 @@ package internal
...
 	"os"
 	"strconv"
+	"strings"
 
-	toml "github.com/pelletier/go-toml/v2"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/rs/zerolog/log"
 )
 
-var COMMENT_PREFIX = "//"
```
jsquyres, line 17: Is there a reason to remove this?

internal/lock.go
```diff
@@ -26,178 +28,295 @@ type config struct {
...
+		Strs("tags", tags).
+		Str("filename", filename).
+		Str("cache_uri", cacheUri).
+		Msg("Adding new resource")
+
+	r, err := NewResourceFromUrlWithCache(paths, algo, tags, filename, cacheUri)
 	if err != nil {
-		return err
+		log.Error().Err(err).Msg("Failed to create resource")
+		return fmt.Errorf("failed to create resource: %w", err)
```
jsquyres, line 81: I'll stop making this same comment after this one: is the error message returned from the failing function insufficient?  If so, should we fix the error message there?

internal/resource.go
```diff
@@ -50,90 +54,127 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
 		Header("Accept", "*/*").
-		ToFile(fileName).
-		Fetch(ctx)
+		ToFile(fileName)
+
+	// Add authorization header if the URL matches the CacheUri and ARTIFACTORY_TOKEN is set.
+	token := os.Getenv("ARTIFACTORY_TOKEN")
+	if token != "" {
+		authHeader := "Bearer " + token
+		req = req.Header("Authorization", authHeader)
```
jsquyres, line 66: I don't think you want to add this to *every* Grabit download -- you only want to add it to cache downloads.

Otherwise, you're sending your Artifactory credential string to the web server where you're downloading from!

internal/resource.go
```diff
@@ -50,90 +54,127 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
 	for _, u := range l.Urls {
-		// Download file in the target directory so that the call to
-		// os.Rename is atomic.
+		log.Debug().Str("URL", u).Msg("Attempting download from URL")
 		lpath, err := GetUrlToDir(u, dir, ctx)
 		if err != nil {
-			downloadError = err
-			break
+			log.Error().Err(err).Str("URL", u).Msg("Download failed, trying next URL")
+			continue
```
jsquyres, line 148: This was just fixed on PR #29

internal/resource.go
```diff
@@ -50,90 +54,127 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
...
-		return err
+
+		return nil // Successful download and integrity check
 	}
-	return nil
+
+	return fmt.Errorf("no valid URLs to download")
 }
 
+// END: Integration of getAuthToken and downloadWithToken logic in Download function
```
jsquyres, line 175: This is a spurious comment; go ahead and remove it.

jsquyres (changes requested):
There appears to be quite a bit of functionality and many changes that are not directly related to the Artifactory / caching functionality, which makes this PR a bit hard to read and review.

Also, can we get the writeup of what we decided would be the first cut of this functionality?  Without that, @rabadin can't properly evaluate this code.

I assume that you notice that the CI tests are failing, too.  You'll want to `make test` before you push again and make sure everything passes.

Finally, let's talk in the next meeting about how to add some unit tests for this functionality.

cmd/add.go
```diff
@@ -1,129 +1,155 @@
...
 	"github.com/spf13/cobra"
 )
 
 func addAdd(cmd *cobra.Command) {
 	addCmd := &cobra.Command{
-		Use:   "add",
+		Use:   "add [url]",
 		Short: "Add new resource",
-		Args:  cobra.MinimumNArgs(1),
+		Args:  cobra.ExactArgs(1),
```
jsquyres, line 22: This doesn't seem right -- it's permissable to add multiple URL's for a resource.

Regardless, if you have additional bug fixes that are unrelated to the Artifactory functionality, please put that stuff in a different PR.  This PR should *only* contain the Artifactory functionality.

cmd/add.go
```diff
@@ -1,54 +1,155 @@
...
 func runAdd(cmd *cobra.Command, args []string) error {
+	// Get flags
 	lockFile, err := cmd.Flags().GetString("lock-file")
 	if err != nil {
-		return err
-	}
-	lock, err := internal.NewLock(lockFile, true)
-	if err != nil {
-		return err
+		return fmt.Errorf("failed to get lock-file flag: %w", err)
```
jsquyres, line 38: Why are you adding all these `fmt.Errorf()` statements?  These do not have anything to do with the functionality of this PR.

If you'd like to propose changing this functionality, great -- do so!  But do it in a different PR.

Please remove all these types of changes from this PR.  This PR should *only* contain the functionality for uploading to / downloading from Artifactory.

cmd/add.go
```diff
@@ -1,54 +1,155 @@
...
+
+	// Check if resource is local file
+	isLocal := !strings.HasPrefix(args[0], "http://") && !strings.HasPrefix(args[0], "https://")
+	if isLocal && cache != "" {
+		log.Debug().
+			Str("file", args[0]).
+			Str("cache", cache).
+			Msg("Uploading local file to Artifactory")
+
+		if err := uploadToArtifactory(args[0], cache); err != nil {
```
jsquyres, line 87: I don't think the file has actually been downloaded yet, has it?

Also, this is an interesting meta question -- I don't think we talked about this case last week: during `grabit add --cache ...`, should grabit upload the file to the cache?  Or does grabit upload the file to the cache the first time you run `grabit download`?

cmd/add.go
```diff
@@ -1,54 +1,155 @@
...
+	client := &http.Client{}
+	resp, err := client.Do(req)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to execute request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Check response
+	if resp.StatusCode != http.StatusCreated && resp.StatusCode != http.StatusOK {
```
jsquyres, line 144: I'm not sure the logic test is right here.

Don't both of these need to be true?

jsquyres:
Note that this PR now has a (simple) conflict that has to be fixed.  Github won't run the CI on a PR that has conflicts, which is why there's currently no CI results here.

cmd/add.go
```diff
@@ -1,54 +1,160 @@
...
+
+	// Add resource to lock file
+	if err := lock.AddResourceWithCache(args, algo, tags, filename, cachePath); err != nil {
+		return fmt.Errorf("failed to add resource to lock file: %w", err)
+	}
+
+	return lock.Save()
+}
+
+func uploadToArtifactory(filePath, cacheUrl string) error {
```
jsquyres, line 122: This function is only called from one location, and that location already had done the `os.Getenv()` to get the token env variable.  It might be better to pass in the token as a param, rather than `os.Getenv()`'ing again.

internal/resource.go
```diff
@@ -1,140 +1,265 @@
...
+			}
+		}
+	}
+
+	if err := r.downloadFromSource(dir, mode, ctx); err != nil {
+		return err
+	}
+
+	// Upload to cache if configured
+	if r.CacheUri != "" && os.Getenv("NO_CACHE_UPLOAD") == "" {
```
jsquyres, line 128: This env variable should probably be prefixed with `GRABIT_`.

internal/resource.go
```diff
@@ -1,140 +1,265 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Resource) downloadFromCache(dir string, mode os.FileMode, ctx context.Context) error {
+	token := os.Getenv("GRABIT_ARTIFACTORY_TOKEN")
+	if token == "" {
+		return fmt.Errorf("GRABIT_ARTIFACTORY_TOKEN must be set for cache operations")
```
jsquyres, line 143: I don't think this is true.  See https://cardmaillouisville-my.sharepoint.com/:w:/g/personal/amdurh05_louisville_edu/ESdWGRZSS8xLuZ17je4MYrgBzV8s02ZNUPRyEtIMoGuc-A?rtime=TVqt2VMD3Ug

internal/resource.go
```diff
@@ -1,140 +1,265 @@
...
+	}
+
+	log.Debug().Str("cache", r.CacheUri).Msg("Attempting cache download")
+	tempFile, err := downloadWithToken(r.CacheUri, dir, ctx, token)
+	if err != nil {
+		return err
+	}
+
+	if err := checkIntegrityFromFile(tempFile, "sha256", r.Integrity, r.CacheUri); err != nil {
```
jsquyres, line 152: I see `sha256` hard-coded here.  Is the integrity always known to be `sha256`?

internal/resource.go
```diff
@@ -1,140 +1,265 @@
...
 		if err != nil {
-			downloadError = err
+			lastErr = err
+			log.Debug().Err(err).Msg("Download failed, trying next URL")
 			continue
 		}
-		err = checkIntegrityFromFile(lpath, algo, l.Integrity, u)
-		if err != nil {
+
+		if err := checkIntegrityFromFile(tempFile, "sha256", r.Integrity, u); err != nil {
```
jsquyres, line 172: Same question as above about the hard-coding of `sha256`.

internal/resource.go
```diff
@@ -1,140 +1,265 @@
...
+	log.Debug().
+		Str("cachePath", cachePath).
+		Msg("Uploading to cache")
+
+	fileData, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+
+	// Use simple http.NewRequest for more control over the request
```
jsquyres, line 212: What does this comment mean?  "more control" compared to what?

jsquyres (changes requested):
Need to remove the a640.... binary.

FWIW: Using `git add -a` or `git commit -a` is dangerous.  It is much safer to `git add ...list of files...` or `git add -p` to select precisely which things you want to add to the staging area.

internal/resource.go
eeasaa01: Using http.NewRequest instead of higher-level HTTP functions gives more detailed control over the request sent to the Artifactory server. This allows the code to set headers, the request body, and other details directly, which might be needed for Artifactory. This level of control is important to make sure the request is formatted correctly for Artifactory. While higher-level libraries are simpler to use, http.NewRequest gives the flexibility needed to work with Artifactory in this case.

rabadin: The higher-level libraries are not just simpler to use, they are much safer and that's why we should use them. See https://github.com/earthboundkid/requests?tab=readme-ov-file#faqs for details.

Can you name *one thing* that the http.NewRequest can do that github.com/carlmjohnson/requests cannot do? If not, let's use github.com/carlmjohnson/requests.

For setting headers, see https://github.com/earthboundkid/requests/wiki#custom-headers

internal/hash.go
```diff
@@ -1,58 +1,206 @@
...
+	hasher.Write(actual)
+	actualBytes := hasher.Sum(nil)
+	actualB64 := base64.StdEncoding.EncodeToString(actualBytes)
+	actualSRI := fmt.Sprintf("%s-%s", h.algo, actualB64)
+
+	return expectedNorm == actualSRI, nil
+}
+
+// VerifyFileIntegrity verifies a file's integrity
+func (h *Hash) VerifyFileIntegrity(path string, expected string) error {
```
rabadin, line 173: This is unused...?

.gitignore
```diff
@@ -1,4 +1,4 @@
-dist/
-grabit
-grabit.lock
-coverage.out
+.DS_Store
+*.bz2
+cmd/test.txt
+cmd/6ae8a75555209fd6c44157c0aed8016e763ff435a19cf186f76863140143ff72
```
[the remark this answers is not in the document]
jsquyres: Please remove the test artifacts from there.  Using a specific hash may not always be correct (e.g., if the file changes someday).  If you're looking to git ignore files that are generated from `make test`, for example, then it might be a good idea to ensure that all those files are in a specific temporary directory, and then ignore that.  But such a change like that should be in a separate PR.

internal/lock.go
```diff
@@ -50,160 +49,199 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
-					}
-				}
-			}
-			if !hasTag {
-				filteredResources = append(filteredResources, r)
+	}
+	return filtered
+}
+
+func hasAllTags(resourceTags, requiredTags []string) bool {
```
rabadin, line 153: Why are you refactoring this? I'm not totally opposed to it, but I think this PR will be complex enough to *not* do anything other than the task at hand.

internal/lock.go
```diff
@@ -50,160 +49,199 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
 	parsed, err := strconv.ParseUint(perm, 8, 32)
 	if err != nil {
-		return NoFileMode, err
+		return NoFileMode, fmt.Errorf("invalid permission format: %w", err)
 	}
 	return os.FileMode(parsed), nil
 }
 
-// Download gets all the resources in this lock file and moves them to
-// the destination directory.
```
rabadin, line 91: Why are you removing this comment?

internal/lock.go
```diff
@@ -50,160 +49,199 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+
+	removed := len(l.conf.Resource) - len(newResources)
+	l.conf.Resource = newResources
+
+	log.Debug().Int("removed", removed).Msg("Resources deleted")
 }
 
 const NoFileMode = os.FileMode(0)
 
-// strToFileMode converts a string to a os.FileMode.
```
rabadin, line 78: Why are you removing this?

internal/lock.go
```diff
@@ -50,160 +49,199 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
 	return os.FileMode(parsed), nil
 }
 
-// Download gets all the resources in this lock file and moves them to
-// the destination directory.
 func (l *Lock) Download(dir string, tags []string, notags []string, perm string) error {
 	if stat, err := os.Stat(dir); err != nil || !stat.IsDir() {
 		return fmt.Errorf("'%s' is not a directory", dir)
 	}
+
```
rabadin, line 122: Why did you add a new line here?

internal/lock.go
```diff
@@ -50,160 +49,199 @@ func NewLock(path string, newOk bool) (*Lock, error) {
 }
 
 func (l *Lock) AddResource(paths []string, algo string, tags []string, filename string) error {
+	return l.AddResourceWithCache(paths, algo, tags, filename, "")
+}
+
+func (l *Lock) AddResourceWithCache(paths []string, algo string, tags []string, filename string, cacheUri string) error {
```
rabadin, line 55: Instead of adding new methods like this, you can simply have one method and use no cache if cacheUri is the empty string. This will be simpler and keep the "interface" surface small.

cmd/download.go
```diff
@@ -27,29 +29,37 @@ func runFetch(cmd *cobra.Command, args []string) error {
...
 	dir, err := cmd.Flags().GetString("dir")
 	if err != nil {
 		return err
 	}
+
 	tags, err := cmd.Flags().GetStringArray("tag")
 	if err != nil {
 		return err
 	}
+
```
rabadin, line 47: Why all the new lines?

cmd/download.go
```diff
@@ -27,29 +29,37 @@ func runFetch(cmd *cobra.Command, args []string) error {
...
 	}
-	err = lock.Download(dir, tags, notags, perm)
-	if err != nil {
+
+	// Download resources
+	if err := lock.Download(dir, tags, notags, perm); err != nil {
 		return err
 	}
+
+	log.Debug().Msg("Download completed successfully")
```
rabadin, line 63: If you really want this, this log message probably belongs to `lock.Download`

internal/lock.go
```diff
@@ -1,22 +1,21 @@
...
 	"os"
 	"strconv"
 
-	toml "github.com/pelletier/go-toml/v2"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/rs/zerolog/log"
 )
 
-var COMMENT_PREFIX = "//"
+// Keeping COMMENT_PREFIX as requested in review
```
rabadin, line 16: We definitely not want this comment.

internal/lock.go
```diff
@@ -1,22 +1,21 @@
...
+const COMMENT_PREFIX = "//"
 
-// Lock represents a grabit lockfile.
```
rabadin, line 19: Why are you removing this comment?

internal/resource.go
```diff
@@ -1,140 +1,263 @@
...
+
+	return &Resource{
+		Urls:      urls,
+		Integrity: integrity,
+		Tags:      tags,
+		Filename:  filename,
+	}, nil
 }
 
-// getUrl downloads the given resource and returns the path to it.
```
rabadin, line 47: Why are you removing this comment?

internal/resource.go
```diff
@@ -1,140 +1,263 @@
...
-	err = requests.
-		URL(u).
+
+	req := requests.URL(u).
 		Header("Accept", "*/*").
-		ToFile(fileName).
-		Fetch(ctx)
+		ToFile(fileName)
+
+	err := req.Fetch(ctx)
```
rabadin, line 64: But... why are you splitting this into two steps if you're not doing anything else with `req`?

internal/resource.go
```diff
@@ -1,140 +1,263 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func (r *Resource) Download(dir string, mode os.FileMode, ctx context.Context) error {
+	// Try cache first if available
+	if r.CacheUri != "" {
+		token := os.Getenv("GRABIT_ARTIFACTORY_TOKEN")
```
rabadin, line 110: You're repeating the env var in too many places, please make it a constant and use the const everywhere.

rabadin (changes requested):
This looks generally okay... but I really don't understand why you're adding so much changes that are completely unrelated to the task at hand and that, in most case, are not real improvements. See my comments inline.

Also: before this change the coverage is 81.8& and after it is 80.4%... could you improve the coverage of the newly-added code to ensure that the coverage doesn't go down with this change?

jsquyres:
It looks like there were some bad merges that were not fixed properly before committing.  For example, I see a full Open MPI tarball added in 4033f12d2d8ce0c1df01f0c249797fa8f5933295.  Word to the wise: you should avoid using `git add -a` or `git commit -a`.  It is far to easy to commit files that you do not intend.

Additionally, `make test` fails to compile the tests (it looks like there's some git conflict markers that were left in the source files).  This is one reason that the CI is failing.  In general, you should probably ensure that grabit builds properly and the tests are passing before pushing up to github.

internal/lock.go
```diff
@@ -176,6 +204,72 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string)
...
+		mode = 0644
+	}
+
+	// Create a temporary file first
+	tmpFile, err := os.CreateTemp(filepath.Dir(filePath), "download-*")
+	if err != nil {
+		return err
+	}
+	tmpPath := tmpFile.Name()
+	defer os.Remove(tmpPath)
```
rabadin, line 257: You're doing a remove... ad on line 270 you're doing a rename... why?

internal/resource_test.go
```diff
@@ -31,17 +30,16 @@ func TestNewResourceFromUrl(t *testing.T) {
 		{
 			urls:  []string{fmt.Sprintf("http://localhost:%d/test.html", port)},
 			valid: true,
-			res:   Resource{Urls: []string{fmt.Sprintf("http://localhost:%d/test.html", port)}, Integrity: fmt.Sprintf("%s-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE=", algo), Tags: []string{}, Filename: ""},
+			res:   Resource{Urls: []string{fmt.Sprintf("http://localhost:%d/test.html", port)}, Integrity: fmt.Sprintf("%s-vvV+x/U6bUC+tkCngKY5yDvCmsipgW8fxsXG3Nk8RyE=", algo), Tags: []string{}, Filename: "", CacheUri: "cacheUri"},
```
rabadin, line 33: Shouldn't we pass the empty string here?

internal/resource_test.go
```diff
@@ -31,17 +30,16 @@ func TestNewResourceFromUrl(t *testing.T) {
...
 		{
 			urls:          []string{"invalid url"},
 			valid:         false,
 			errorContains: "failed to download",
 		},
 	}
-
 	for _, data := range tests {
-		resource, err := NewResourceFromUrl(data.urls, algo, []string{}, "")
+		resource, err := NewResourceFromUrl(data.urls, algo, []string{}, "", "cacheUri")
```
rabadin, line 42: And here?

internal/resource.go
```diff
@@ -122,8 +141,12 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 		ok = true
 	}
 	if !ok {
-		if downloadError != nil {
-			return downloadError
+		if err == nil {
```
rabadin, line 144: Why are you changing this?

internal/resource.go
```diff
@@ -95,7 +114,7 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 		lpath, err := GetUrlToDir(u, dir, ctx)
 		if err != nil {
 			downloadError = err
-			continue
```
rabadin, line 98: Why this change, can you explain it?

internal/resource.go
```diff
@@ -83,6 +84,24 @@ func GetUrltoTempFile(u string, ctx context.Context) (string, error) {
...
+			localName = path.Base(l.Urls[0])
+		}
+		resPath := filepath.Join(dir, localName)
+
+		err := downloadFromArtifactory(ctx, l.CacheUri, l.Integrity, resPath, mode)
+		if err == nil {
+			return nil
+		}
+		// Return the error and let Lock show the warning
+		return err
```
rabadin, line 102: So... you're failing here if the download from the cache doesn't work. This is no what we said we should do; we should:
- try to download from the cache
- if it works, return success
- if it doesn't work, issue a warning and continue the execution, trying to download the resource from the other urls

internal/lock_test.go
```diff
@@ -51,7 +51,7 @@ func TestLockManipulations(t *testing.T) {
 	port, server := test.HttpHandler(handler)
 	defer server.Close()
 	resource := fmt.Sprintf("http://localhost:%d/test2.html", port)
-	err = lock.AddResource([]string{resource}, "sha512", []string{}, "")
+	err = lock.AddResource([]string{resource}, "sha512", []string{}, "", "cacheUri")
```
rabadin, line 54: Why is it "cacheUri" and not the empty string?

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,207 @@
...
+		port, server := test.HttpHandler(handler)
+		defer server.Close()
+
+		err := runDownloadWithCache(fmt.Sprintf("http://localhost:%d/cache", port))
+		assert.NoError(t, err) // Should work without an error
+	})
+}
+
+// Simulated download
+func runDownloadWithCache(cacheURL string) error {
```
rabadin, line 158: I don't understand what you're doing here. You're just downloading something using the standard library... and this is what you're testing? This makes little sense to me... can you explain why you're doing this?
The name seems to imply that you're exercising grabit's Download method but you're not doing this.

rabadin (changes requested):

grsbit
jsquyres, line 1: Also need to remove this binary file.

Do not commit binary files to Git.

internal/lock.go
```diff
@@ -49,20 +57,92 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+		return fmt.Errorf("GRABIT_ARTIFACTORY_TOKEN environment variable is not set")
+	}
+
+	// Read file
+	fileContent, err := os.ReadFile(filePath)
+	if err != nil {
+		return fmt.Errorf("failed to read file: %v", err)
+	}
+
+	// Get hash from integrity string
```
rabadin, line 107: Why are you doing all this? The SRI format is perfectly fine to use as is as a filename without doing any transformation to it...

internal/lock.go
```diff
@@ -49,20 +57,92 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+	hashBytes, err := base64.StdEncoding.DecodeString(h)
+	if err != nil {
+		return fmt.Errorf("failed to decode hash: %v", err)
+	}
+
+	hexHash := hex.EncodeToString(hashBytes)
+	artifactoryURL := fmt.Sprintf("%s/%s", cacheURL, hexHash)
+
+	// Create upload request
+	req, err := http.NewRequest("PUT", artifactoryURL, bytes.NewReader(fileContent))
```
rabadin, line 124: As discussed, let's use `github.com/carlmjohnson/requests` for non-test http interactions, like the rest of the code does.

internal/lock.go
amdurh05: @rabadin This should be fixed now.

internal/lock.go
```diff
@@ -49,30 +56,120 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+			if err != nil {
+				slog.Warn("Warning: Unable to delete from Artifcatory")
+			}
 		}
 	}
 	l.conf.Resource = newStatements
 }
 
+func deleteCache(url, token string) error {
+	req, err := http.NewRequest("DELETE", url, nil)
```
rabadin, line 156: Same remark here, let use `github.com/carlmjohnson/requests` for http interactions.

rabadin (changes requested):

rabadin:
@amdurh05  Also, you have a random 11M binary file in here.

cmd/download_test.go
```diff
@@ -112,9 +112,9 @@ func TestRunDownloadFailsIntegrityTest(t *testing.T) {
 	content := `abcdef`
 	port := test.TestHttpHandler(content, t)
 	testfilepath := test.TmpFile(t, fmt.Sprintf(`
-	[[Resource]]
```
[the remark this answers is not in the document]
eeasaa01: We did not make any changes in this part of the file. I believe these whitespace modifications were introduced by Mr. Raphael in a prior commit.

cmd/download_test.go
jsquyres, line 1: This file *is* poorly named, but let's not rename it as part of this PR -- let's rename it in a separate PR.

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+		cacheURL := fmt.Sprintf("http://localhost:%d", port)
+
+		err = lock.AddResource([]string{sourceURL}, "sha256", []string{}, "", cacheURL)
+		assert.Contains(t, err.Error(), "GRABIT_ARTIFACTORY_TOKEN environment variable is not set")
+	})
+}
+func TestDownloadWithArtifactoryCache(t *testing.T) {
+	t.Run("NO_CACHE_UPLOAD", func(t *testing.T) {
+		// Turn on NO_CACHE_UPLOAD setting
+		os.Setenv("NO_CACHE_UPLOAD", "1")
```
rabadin, line 43: Use https://beta.pkg.go.dev/testing@master#T.Setenv instead. This guarantees that it gets restored after the test runs.

rabadin:
@eeasaa01 Btw, you have a conflict that you need to fix by rebasing your work onto the latest main.

rabadin:
@eeasaa01 I suggested doing a rebase. You did a merge. That's going to potentially complicate the process when you squash all your changes into just one commit.

eeasaa01:
I did the merge through the GitHub website, as it was the only option I saw there—apologies for that. Also, I now see the tests failing in two files: lock_test and download_test. What do you recommend to fix these?
________________________________

jsquyres:
> I did the merge through the GitHub website, as it was the only option I saw there—apologies for that. 

In the future: although GitHub does have the option for Rebasing (instead of Merging), you might be better served doing the operation on the command line or in your local IDE -- it might be slightly less confusing as to what is actually happening.

As we talked about in our status meetings, merging the target branch into your PR branch frequently leads to pain in the long run.  It should be avoided whenever possible.  Rebase will generally spare you that pain.

> Also, I now see the tests failing in two files: lock_test and download_test. What do you recommend to fix these?

Your PR removed the 5th argument from `lock.Download()`.
 
https://github.com/cisco-open/grabit/pull/30/files#diff-4d8cb8fc3d173e6b0cd63fafbe1a7ee6641bddcea1fc7e13d6c07b00d1d684c8L92-R184
 
* If you didn't mean to do that, put that 5th argument back.
* If you did mean to do that, then everywhere that calls `lock.Download()` needs to be adjusted to remove that 5th argument.

internal/lock.go
```diff
@@ -150,28 +242,32 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string,
...
-		statusLine = NewStatusLine(ctx, &filteredResources)
-		err := statusLine.InitResourcesSizes()
-		if err == nil {
-			statusLine.Start(true)
-		} else {
-			statusLine = nil // Do not update or display SL.
-		}
-	}
-
-	for i, r := range filteredResources {
```
jsquyres, line 165: Was this code just added in #24?  If so, you shouldn't be removing it.

internal/lock.go
```diff
@@ -150,28 +242,32 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string,
...
+					fmt.Printf("Failed to download from Artifactory, falling back to original URL: %v\n", err)
+				}
+			}
 
 			err := resource.Download(dir, mode, ctx)
 			errorCh <- err
-
-			if statusLine != nil {
-				statusLine.Increment(i)
-			}
```
jsquyres, line 174: Same question as above: if this was just added in #24, you shouldn't be removing it.

jsquyres:
@eeasaa01 and I are DM'ing in Teams to try to clean this up.  I just took a first step here, and did the following:

1. Removed the problematic merge (and all the problematic commits that came after it) from earlier today
2. Squashed all the commits from after 3396a6f into a single commit

We still have a problematic merge in the branch on this commit.  The merge from https://github.com/eeasaa01/grabit/pull/1 contains a problematic merge from main as well, which is still causing more conflicts.  Still need to resolve this...

jsquyres:
Here's what the DAG on this PR looks like right now:

```
* e28f6ce (HEAD -> feature/artifactory-upload, origin/feature/artifactory-upload-SAVE-3-after-squashing, jsq/feature>
*   3396a6f Merge pull request #1 from eeasaa01/feature/artifactory-delete
|\  
| *   d4dcfd7 Merge branch 'feature/artifactory-upload' into feature/artifactory-delete
| |\  
| |/  
|/|   
* | a10e728 Update .gitignore
* | 335070a Update hash.go and lock.go based on code review
* | 5ba7cdd Update log level to warn when failing to upload to cache
* | f2e8c42 Update cmd/cache_test.go with cache download tests
* | 8a41679 Update cmd/add.go with feedback implementation
* | 513e420 Fix: Use requests library for cache upload
* | 5d9a7c8 Fix: Code improvements and cleanup
* | f0676ab Fix: Prefix NO_CACHE_UPLOAD with GRABIT_
* | 1ee74b5 Fix: Pass token as parameter to uploadToArtifactory
* | 2090e45 Fix: Code review feedback and cleanup
* | 2ab9448 Cleanup: Remove binary files and add gitignore
* | c389fb2 Added comprehensive tests for Artifactory cache functionality - Added TestGrabitRequirements - Tested to>
* | 4033f12 Implemented requested changes and added tests for Artifactory integration
* | 410b85b Fixed Artifactory integration and addressed PR feedback
* | f5f034e Restored COMMENT_PREFIX and improved error logging for resource creation failures
* | 503e529 Conditional ARTIFACTORY_TOKEN check and removed spurious comment
* | e98d7af Fixed authentication error and adjusted log level to debug for successful uploads
* | 61ec037 Implemented Artifactory upload and download handling
* | 7936916 Try all resources URLs before failing (#29)
| * a484aa3 Updating delete function to also delete artifactory cache
| * 082c98b Added comprehensive tests for Artifactory cache functionality - Added TestGrabitRequirements - Tested to>
| * 0ea343f Implemented requested changes and added tests for Artifactory integration
| * d6229c9 Fixed Artifactory integration and addressed PR feedback
| * 825c3ea Restored COMMENT_PREFIX and improved error logging for resource creation failures
| * 412a3aa Conditional ARTIFACTORY_TOKEN check and removed spurious comment
| * 0a01f2e Fixed authentication error and adjusted log level to debug for successful uploads
| * 1675df4 Implemented Artifactory upload and download handling
|/  
* ba3ac86 Add more unit-tests (#22)
*   b350ab9 (tag: v1.0.3) Merge pull request #21 from rabadin/cleanup-tmp-dir
```

Something apparently went wrong -- there was apparently some kind of drift between @amdurh05's branch and this branch when https://github.com/eeasaa01/grabit/pull/1 was merged.

One way to resolve this -- which will be somewhat painful -- is:

```bash
# Check out the correct branch
git checkout feature/artifactory-upload

# Save this location in the DAG, just in case you want to go back / try again
git checkout -b feature/artifactory-upload-SAVE

# Go back to the original branch
git checkout feature/artifactory-upload

# Reset this branch + working tree to be before the problematic merge
git reset --hard a10e728

# Cherry pick the 1 commit that we want from https://github.com/eeasaa01/grabit/pull/1
# (i.e., the commit from @amdurh05)
git cherry-pick a484aa3

# Now cherry pick the 1 remaining commit that was originally 
# on this PR branch -- i.e., the commit that is at feature/artifactory-upload-SAVE
# (i.e., the commit that is the squash of all the commits that were after the
# problematic merge)
git cherry-pick e28f6ce
```

This 2nd cherry pick will generate a lot of conflicts that you will need to resolve.

Once you resolve these conflicts and commit the result, then you will need to rebase -- **NOT MERGE!** 😄 -- on top of `origin/main` to get all of @MDylanWilliams's progress bar changes.

@rabadin Is there a better / easier way?

.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
         go-version-file: 'go.mod'
 
     - name: Test
+      env:
+        GRABIT_ARTIFACTORY_TOKEN: team2grabit
```
rabadin, line 23: I don't think that's a good idea. You want to control this in code via https://pkg.go.dev/testing#B.Setenv. This way you can also test what happens when it's *not* set.

rabadin (changes requested):

.github/workflows/ci.yml
eeasaa01: I thought this way it may work, but okay, I will use your suggestion right away.

cmd/download_test.go
jsquyres: Let's not forget to undo this change, too.

internal/lock.go
```diff
@@ -49,30 +56,110 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+				log.Warn().Msg("Warning: Unable to delete from Artifcatory: GRABIT_ARTIFACTORY_TOKEN not set.")
+
+				continue
+			}
+
+			artifactoryURL := fmt.Sprintf("%s/%s", r.CacheUri, r.Integrity)
+
+			err := deleteCache(artifactoryURL, token)
+			if err != nil {
+				log.Warn().Msg("Warning: Unable to delete from Artifcatory")
```
rabadin, line 141: Typo: Artifcatory

internal/lock.go
rabadin: Yes, this looks like the result of a bad conflict resolution. You shouldn't be removing this code.

internal/lock.go
```diff
@@ -89,7 +176,7 @@ func strToFileMode(perm string) (os.FileMode, error) {
 
 // Download gets all the resources in this lock file and moves them to
 // the destination directory.
-func (l *Lock) Download(dir string, tags []string, notags []string, perm string, status bool) error {
+func (l *Lock) Download(dir string, tags []string, notags []string, perm string, someBool bool) error {
```
rabadin, line 179: What's this change, why the name change?

internal/lock.go
rabadin: You have resolved this without giving me a good explanation... re-opening...

rabadin (changes requested):

internal/lock.go
eeasaa01: I was to trying to make sure the temporary file is deleted if something goes wrong. If everything works fine, the file get renamed and moved to where it needs to be.
Is there a better way to do this?

internal/lock.go
eeasaa01: I didn’t really change it. When I was testing, the function needed five arguments, so I added a temporary one called someBool. Later, when I added Dylan’s code, I saw it should be status, so I fixed it.

internal/lock.go
rabadin: Yes: just remove all this code and re-use the code that exists; you only need to add support for a bearer token:
```
diff --git a/internal/resource.go b/internal/resource.go
index a377e87..aab11b0 100644
--- a/internal/resource.go
+++ b/internal/resource.go
@@ -45,17 +45,20 @@ func NewResourceFromUrl(urls []string, algo string, tags []string, filename stri
 }

 // getUrl downloads the given resource and returns the path to it.
-func getUrl(u string, fileName string, ctx context.Context) (string, error) {
+func getUrl(u string, fileName string, bearer string, ctx context.Context) (string, error) {
 	_, err := url.Parse(u)
 	if err != nil {
 		return "", fmt.Errorf("invalid url '%s': %s", u, err)
 	}
 	log.Debug().Str("URL", u).Msg("Downloading")
-	err = requests.
+	req := requests.
 		URL(u).
 		Header("Accept", "*/*").
-		ToFile(fileName).
-		Fetch(ctx)
+		ToFile(fileName)
+	if bearer != "" {
+		req.Header("Authorization", fmt.Sprintf("Bearer %s", bearer))
+	}
+	err = req.Fetch(ctx)
 	if err != nil {
 		return "", fmt.Errorf("failed to download '%s': %s", u, err)
 	}
@@ -64,12 +67,12 @@ func getUrl(u string, fileName string, ctx context.Context) (string, error) {
 }

 // GetUrlToDir downloads the given resource to the given directory and returns the path to it.
-func GetUrlToDir(u string, targetDir string, ctx context.Context) (string, error) {
+func GetUrlToDir(u string, targetDir string, bearer string, ctx context.Context) (string, error) {
 	// create temporary name in the target directory.
 	h := sha256.New()
 	h.Write([]byte(u))
 	fileName := filepath.Join(targetDir, fmt.Sprintf(".%s", hex.EncodeToString(h.Sum(nil))))
-	return getUrl(u, fileName, ctx)
+	return getUrl(u, fileName, bearer, ctx)
 }

 // GetUrlWithDir downloads the given resource to a temporary file and returns the path to it.
@@ -79,7 +82,7 @@ func GetUrltoTempFile(u string, ctx context.Context) (string, error) {
 		log.Fatal().Err(err)
 	}
 	fileName := file.Name()
-	return getUrl(u, fileName, ctx)
+	return getUrl(u, fileName, "", ctx)
 }

 func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) error {
@@ -92,7 +95,7 @@ func (l *Resource) Download(dir string, mode os.FileMode, ctx context.Context) e
 	for _, u := range l.Urls {
 		// Download file in the target directory so that the call to
 		// os.Rename is atomic.
-		lpath, err := GetUrlToDir(u, dir, ctx)
+		lpath, err := GetUrlToDir(u, dir, "", ctx)
 		if err != nil {
 			downloadError = err
 			continue
```

rabadin (changes requested):

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+		err = lock.AddResource([]string{sourceURL}, "sha256", []string{}, "", cacheURL)
+		// Verify that the error message indicates the token is not set.
+		assert.Contains(t, err.Error(), "GRABIT_ARTIFACTORY_TOKEN environment variable is not set")
+	})
+}
+
+// TestDownloadWithArtifactoryCache verifies downloading resources with caching enabled.
+func TestDownloadWithArtifactoryCache(t *testing.T) {
+	// Sub-test to verify behavior when NO_CACHE_UPLOAD is set.
+	t.Run("NO_CACHE_UPLOAD", func(t *testing.T) {
```
rabadin, line 49: We discussed this: there is no need for `NO_CACHE_UPLOAD`. So this test can go.

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/cisco-open/grabit/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestAddWithArtifactoryCache verifies adding a resource with caching enabled.
+func TestAddWithArtifactoryCache(t *testing.T) {
+	// Sub-test to verify behavior when the token is not set.
+	t.Run("TokenNotSet", func(t *testing.T) {
```
rabadin, line 18: What's the benefit of using `t.Run` if you only have one `t.Run` per test?

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+		// Clear the GRABIT_ARTIFACTORY_TOKEN environment variable.
+		t.Setenv("GRABIT_ARTIFACTORY_TOKEN", "")
+
+		// Setup a simple HTTP handler that always returns "test content".
```
rabadin, line 22: Is all this setup really necessary if you're only testing the case where GRABIT_ARTIFACTORY_TOKEN isn't set?

internal/artifactory_cache_test.go
```diff
@@ -0,0 +1,139 @@
...
+		lock.DeleteResource(sourceURL)
+
+		// Save the lock file again after deletion.
+		err = lock.Save()
+		require.NoError(t, err) // Fail the test if saving fails.
+
+		// Reload the lock file and verify the resource is gone.
+		newLock, err := NewLock(lockPath, false)
+		require.NoError(t, err)
+		assert.Equal(t, 0, len(newLock.conf.Resource)) // Ensure no resources remain.
```
rabadin, line 137: Here you're just testing that the resource deleting works... but that's not at all what the new code does... so why are you testing this at all?

internal/lock.go
```diff
@@ -49,30 +55,111 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+	// If cache URL is provided, handles Artifactory upload
+	if cacheURL != "" {
+		token := os.Getenv("GRABIT_ARTIFACTORY_TOKEN")
+		if token == "" {
+			return fmt.Errorf("GRABIT_ARTIFACTORY_TOKEN environment variable is not set")
+		}
+
+		// Add context here
+		ctx := context.Background()
+		path, err := GetUrltoTempFile(paths[0], "", ctx) // Add empty string for bearer token
```
rabadin, line 78: Isn't it a bit weird to pass the empty string here?

internal/lock.go
```diff
@@ -49,30 +55,111 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
 }
 
+func uploadToArtifactory(filePath, cacheURL, integrity string) error {
+	token := os.Getenv("GRABIT_ARTIFACTORY_TOKEN")
+	if token == "" {
+		return fmt.Errorf("GRABIT_ARTIFACTORY_TOKEN environment variable is not set")
+	}
+
+	// Read the content of the file
+	fileContent, err := os.ReadFile(filePath)
```
rabadin, line 102: You don't need to read the file yourself, you can just use `BodyFile(filePath)`.

internal/lock.go
```diff
@@ -49,30 +55,111 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
+	if err != nil {
+		return fmt.Errorf("failed to read file: %v", err)
+	}
+
+	// Use the integrity value directly for the URL since it's already in the correct format
+	artifactoryURL := fmt.Sprintf("%s/%s", cacheURL, integrity)
```
rabadin, line 108: This is repeated in here and in DeleteResource and this is bad since these 2 calls absolutely need to be in sync. You need to refactor this into a utility method called in both places.

internal/lock.go
```diff
@@ -49,30 +55,111 @@ func NewLock(path string, newOk bool) (*Lock, error) {
...
 
+func deleteCache(url, token string) error {
+	// Create and send a DELETE request with an Authorization header.
+	err := requests.
+		URL(url).                                                 // Target URL for the DELETE request.
+		Method(http.MethodDelete).                                // Set HTTP method to DELETE.
+		Header("Authorization", fmt.Sprintf("Bearer %s", token)). // Add token in the Authorization header.
+		Fetch(context.Background())                               // Execute the request.
+
+	if err != nil {
```
rabadin, line 156: No need for this, you can just do:
```
return requests.URL(url).[...]
```

internal/lock_test.go
```diff
@@ -33,7 +33,6 @@ func TestNewLockValid(t *testing.T) {
 	assert.Equal(t, "sha256-asdasdasd", statement.Integrity)
 	assert.Equal(t, []string{"tag1", "tag2"}, statement.Tags)
 }
-
```
rabadin, line 36: Why the line removal?

internal/lock.go
```diff
@@ -165,6 +252,36 @@ func (l *Lock) Download(dir string, tags []string, notags []string, perm string,
...
+							errorCh <- nil
+							if statusLine != nil {
+								statusLine.Increment(i)
+							}
+							return
+						}
+					}
+					if strings.Contains(err.Error(), "lookup invalid") || strings.Contains(err.Error(), "dial tcp") {
+						fmt.Printf("Failed to download from Artifactory, falling back to original URL: %v\n", err)
+					}
```
rabadin, line 282: I might be missing something here... but where are you doing the file integrity check?

rabadin (changes requested):

jsquyres (dismissed):
It doesn't look like the progress bar code was re-integrated properly.  When I run:

```
$ ./grabit download --status
✔[████████████████████]          3/3 Resources          126 MB / 126 MB          13s elapsed 
```

I see an empty status bar for 13 seconds (and 0/3 Resources  0 B / 126 MB) and then all the sudden it's full / done.  Something is not right -- it should be continually updating with its actual status.

jsquyres:
@MDylanWilliams I think this was a bad test on my part.  With more proper testing on my side, I think you're right -- it *is* working properly.  Sorry for the confusion!

internal/lock_test.go
eeasaa01: I removed it accidentally. I’ll redo that, sorry!

internal/lock.go
eeasaa01: I will change it.

internal/lock.go
eeasaa01: working on address that

internal/lock.go
eeasaa01: got it!

internal/artifactory_cache_test.go
eeasaa01: I think the test cases haven't been updated to align with the new code yet. Do you have please any suggestions that might help with updating the test cases to better reflect the changes?

internal/resource_test.go
```diff
@@ -1,6 +1,5 @@
 // Copyright (c) 2023 Cisco Systems, Inc. and its affiliates
 // All rights reserved.
-
```
jsquyres, line 3: FWIW: here's another spurious line removal.

Keep in mind: there's not a technical problem with removing this blank line.

But it's an extra change that has no purpose in this PR.  So you should probably put that blank line back, just so that this PR **ONLY** has the Artifactory changes -- no spurious changes.

There's at least 2 other places where blank lines were deleted from the tops of `.go` files.  Please put them all back.  Thanks!

internal/artifactory_cache_test.go
amdurh05: @rabadin Looks like without defining "t.Setenv("GRABIT_ARTIFACTORY_TOKEN", "")" the test fails when I run "make test" but passes when I run the test directly. Do you have any recommendations for this? 

Went ahead and pushed this change to see if it passes the CI.

rabadin: I'm not sure what you mean by "when I run the test directly". Do you mean running this individual test in isolation or manually testing things?
I see that the tests pass right but you're not testing all the functionality, which is reflected in the new coverage number (with this PR): 75.8%.

amdurh05: @rabadin I mean when I run the test in isolation. 

I think one way to up the testing coverage could be testing the normal Artifactory download functionality, I don't think this is currently being tested. If you think this would be a good test to add I can work on adding it when I get home this evening.

rabadin: By definition the UT coverage is the coverage of the UT.

rabadin: Well, yes, you clearly need to test:
a) the happy path
- Adding a resources with a cache
- Downloading a resource with a cache, getting the data from the cache (using mocking, as discussed)
- Deleting a resource with a cache, deleting the cache (using mocking, as discussed)
b) the non happy path
- Adding resource with a cache but no token env var
- Downloading a resource with a cache, but the cache GET errors
- Deleting a resource with a cache, and deleting the cache works

jsquyres (approved):

rabadin (approved):

